PR #10, opened 2019-01-24
Updated prettier.

mhaertwig:
- Newer version.
- Displaying errors using eslint-plugin-prettier


.eslintrc.json
```diff
@@ -7,11 +7,13 @@
...
   "parser": "babel-eslint",
   "extends": ["airbnb", "prettier"],
+  "plugins": ["prettier"],
   "rules": {
     "import/no-extraneous-dependencies": ["error", {"devDependencies": true }],
     "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
     "react/forbid-prop-types": "Off",
     "max-len": ["error", { "code": 80, "ignoreComments": true }],
-    "react/jsx-curly-spacing": ["error", { "when": "never" }]
+    "react/jsx-curly-spacing": ["error", { "when": "never" }],
```
oterral: you can remove max-len and react/jsx-curly-spacing we don't need it anymore, prettier takes care of it

oterral:
@mhaertwig what do you think remove the `.prettierrc.json`  file if if we adopt it, we might as well adopt it fully.

mhaertwig:
@oterral Nooo. The semicolon thing would be too much for me.

mhaertwig:
The code would be a mess without prettier! I'm lucky that we have it.

oterral:
You just join a big and happy family !!!

PR #11, opened 2019-01-25
Simplify examples

mhaertwig:


src/components/autocomplete/Autocomplete.js
```diff
@@ -57,17 +57,18 @@ const propTypes = {
   /**
    * Get the key for each item used when react creates the list.
    */
-  getItemKey: PropTypes.func.isRequired,
+  getItemKey: PropTypes.func,
```
oterral: it was required because React complains when a key is not specified in a for loop.

src/components/link/ActionLink.js
```diff
@@ -13,7 +13,7 @@ const propTypes = {
   /**
    * Title attribute.
    */
-  title: PropTypes.string.isRequired,
+  title: PropTypes.string,
```
oterral: I think I put it 'required' because, for the  "barrierefreiheit", we HAVWE TO put a title  and ariaLabel everywhere, that avoid the devloper to forget about this.
I think we should discuss this kind of things on monday.

PR #12, opened 2019-01-29
Improve styleguide

oterral:


friedjoff:
The `styleguide` directory gets recreated each time we run `yarn doc` so I guess we can delete it from the repository. Looks good otherwise!

PR #13, opened 2019-01-30
Improve autocomplete

oterral:
- Add onFocus and onBlur event on Autocomplete comp
- Close results list when we click on search button


mhaertwig:
I think I would have fixed this outside the component but your idea is much better.

oterral:
It was more tricky than I thought . This way is a bit cleaner.

PR #14, opened 2019-01-31
[TRAFSHUL-6] Create generic WMTS Layer

edellucien:
[TRAFSHUL-6] Create generic WMTS Layer

### DIFF
--- a/src/WMTSLayer.js
+++ b/src/WMTSLayer.js
@@ -1,0 +1,40 @@
+import TileLayer from 'ol/layer/Tile';
+import WMTSSource from 'ol/source/WMTS';
+import WMTSTilegrid from 'ol/tilegrid/WMTS';
+import Layer from './Layer';
+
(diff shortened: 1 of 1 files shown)


friedjoff:
I agree with @mhaertwig. I guess we should only create new layer classes for layers that require custom methods.

oterral:
hmm what kind of custom methods ?  on Layer and VectorLayer I only see shortcuts methods. I think it's confusing to know what we must use. On interactive-stationplans we use the `layer.olLayer`  directly  on a lot places. So if we can't use this class to avoid duplicate code through applications (like the wmtsResolutions) I think these Classes are useless.

mhaertwig:
I'm not sure if I'm getting you right. I see the _wmtsResolutions_ as service/application specific. Implementing a class to avoid duplicate code should happen inside the app, not the library. Btw.: For me the _VectorLayer_ in its current state should be removed from _react-spatial_ as well. After the last changes it does not provide any additional value. What do you think?

PR #16, opened 2019-02-05
Convert 'mouseup' event to 'click'

edellucien:
fix safari popup click event which targets the map instead
https://github.com/openlayers/openlayers/issues/6948


oterral:
@mhaertwig  For me it's ok if  we merge this for golive  then it will be replaced by  #19 after golive.

mhaertwig:
@oterral when do you think is #19 ready to be merged?

oterral:
@mhaertwig  now , I just finished to test the BarrierFree in interactive.stationplans

PR #19, opened 2019-02-07
Replace use of ol.Overlay.

oterral:
Long-term alternative to #16.


mhaertwig:
I like your solution. Thanks for implementing it.

PR #23, opened 2019-02-12
Add LayerTree

oterral:
@edellucien


mhaertwig:
Looks nice so far! What about centralizing the layer interaction behavior in a `LayerService` (see spec). Is it still planned?

oterral:
of course we need to discuss that. @edellucien begins the work on this.For now, the layer Tree is totally agnostic from  the map. And we should keep that way if we can, it will better for maintenance.

oterral:
Ready to review. I will add the stuff to have a configurable css theme in another PR.

src/components/layertree/LayerTree.js
```diff
@@ -0,0 +1,389 @@
...
+          isExpanded: PropTypes.bool,
+        }),
+      }),
+    ).isRequired,
+  }),
+
+  /**
+   * CSS class to apply on each item.
+   */
+  classNameItem: PropTypes.string,
```
mhaertwig: For consistency (see e.g. Button.js) please use `className`

mhaertwig:
Nice job so far. Thanks! Some comments:
- The behavior in barrier free mode should be the same as on https://maps.trafimage.ch
- Please make the example a bit nicer:
  - Show layer tree on map
  - Use some better example layers with their real names
  - The interactive stationplan layer isn't really a good example layer. Please use something else (e.g. OSM).
  - Remove the Japanese flag from the examples.

src/components/layertree/LayerTree.js
oterral: `className`, it  feels  it's for the parent component.  but it's for the item.

oterral:
@mhaertwig  for better examples, @edellucien is on it with the LayerService.

oterral:
I close this in favor of https://github.com/geops/react-spatial/pull/30

PR #25, opened 2019-02-19
Fix radio click event: should not uncheck checkbox on the same node, should uncheck radio siblings children

edellucien:
Fix radio click event, not unchecking checkbox on the same node


src/components/layertree/LayerTree.js
```diff
@@ -242,7 +275,7 @@ class LayerTree extends PureComponent {
   /**
    * Apply a mutation to all the children recursively.
    */
-  applyToChildren(tree, item, mutation, itemIgnored) {
+  applyToChildren(tree, item, mutation, isIgnored) {
```
oterral: Call it `isIgnoredFunc` , otherwise we have the feeling that the parameter is a boolean.

src/components/layertree/LayerTree.js
```diff
@@ -251,7 +284,7 @@ class LayerTree extends PureComponent {
     tree.items[item.id].children.forEach(childId => {
       const child = newTree.items[childId];
 
-      if (itemIgnored && child.id === itemIgnored.id) {
+      if (isIgnored !== undefined && isIgnored(child)) {
```
oterral: You can just test if it exists:
```
if (isIgnored && isIgnored(child))
```

src/components/layertree/LayerTree.test.js
```diff
@@ -124,51 +133,77 @@ describe('LayerTree', () => {
...
-          '3': {
-            isChecked: true,
-            isExpanded: true,
-            type: 'radio',
-          },
-          '1-1': {
-            defaults: {
+    test(
+      'expands the item, uncheck and collapse all the siblings and their ' +
+        'radio siblings children, and apply defaults value to the children.',
```
oterral: expands the item, uncheck and collapse all the radio siblings and their children, then apply default to the item's children ...

src/components/layertree/LayerTree.js
```diff
@@ -281,12 +314,7 @@ class LayerTree extends PureComponent {
       }
 
       if (child.hasChildren) {
-        newTree = this.applyToChildren(
-          newTree,
-          child,
-          newMutation,
-          itemIgnored,
-        );
+        newTree = this.applyToChildren(newTree, child, newMutation);
```
oterral: No more use of ` itemIgnored` ?

src/components/layertree/LayerTree.js
```diff
@@ -152,12 +152,8 @@ class LayerTree extends PureComponent {
...
             isExpanded: !value,
           },
-          child => {
-            if (child.id === item.id || child.type === 'checkbox') {
-              return true;
-            }
-            return false;
-          },
+          child => child.id === item.id,
+          child => child.type === 'checkbox',
```
oterral: Why 2 functions now? one is enough no?

edellucien: It is aimed to apply 'isIgnoredFunc' when re-applying to children, but without the filter on checkbox.

src/components/layertree/LayerTree.js
```diff
@@ -242,7 +271,7 @@ class LayerTree extends PureComponent {
   /**
    * Apply a mutation to all the children recursively.
    */
-  applyToChildren(tree, item, mutation, itemIgnored) {
+  applyToChildren(tree, item, mutation, isIgnoredFunc, isCheckboxFunc) {
```
oterral: I understand better now maybe a function named `isTypeIgnoredFunction` makes more sense.

edellucien: I apply this naming in the next MR.

oterral: I ve done it directly on github.

PR #30, opened 2019-02-25
LayerTree using layer events

oterral:
Here a first try using layer's events. 

Config is read by the `ConfigReader`  class, this class create layers and pass it to the `LayerService`.
Then `LayerTree`  listen custom  LayerServcie's `change:visible`  events form the LayerService' and try to update other layers accordingly .


oterral:
Ok now it's reviewable, all seems to work.

PR #33, opened 2019-02-27
Simplifying code.

mhaertwig:
Simplifying the code of the `LayerService`


src/components/layertree/LayerTree.js
```diff
@@ -217,42 +207,38 @@ class LayerTree extends Component {
         tabIndex={tabIndex}
         className={classNameToggle}
         onClick={() => {
-          if (!level || !LayerTree.hasChildren(item)) {
-            LayerTree.onInputClick(item);
+          if (layer.getChildren().length === 0) {
+            layer.setVisible(!layer.getVisible());
```
oterral: That's why onInputClick was for , to avoid duplicating this code, so you don't have to test it on each case.

mhaertwig: That makes sense. Sorry, I removed it for testing and forgot to re-add it.

src/Layer.js
```diff
@@ -78,17 +89,54 @@ export default class Layer extends Observable {
...
+    return !!this.children.find(l => l.getVisible());
+  }
+
+  hasChildren(visible) {
+    return !!this.children.find(l => visible === l.getVisible());
   }
 
   getProperties() {
-    return this.olLayer.getProperties();
+    if (this.olLayer) {
```
mhaertwig: Memo: This can be removed as well.

src/components/layertree/LayerTree.js
```diff
@@ -109,106 +136,77 @@ class LayerTree extends Component {
...
           type={inputType}
-          name={item.getRadioGroup()}
+          name={layer.getRadioGroup()}
           tabIndex={-1}
-          checked={item.getVisible()}
+          checked={layer.getVisible()}
           onChange={() => {}}
           onClick={() => {
-            LayerTree.onInputClick(item);
+            this.onInputClick(layer, layer.getChildren().lenght);
```
oterral: `length`

src/components/layertree/LayerTree.js
```diff
@@ -109,106 +136,77 @@ class LayerTree extends Component {
...
     );
   }
 
-  renderArrow(item) {
+  renderArrow(layer) {
     const { classNameArrow } = this.props;
-    if (!LayerTree.hasChildren(item)) {
+    const { expandedLayerNames } = this.state;
+
+    if (layer.getChildren().length === 0) {
```
oterral: No need to compare to 0

src/components/layertree/LayerTree.js
```diff
@@ -109,106 +136,77 @@ class LayerTree extends Component {
...
       return null;
     }
+
     return (
       <div
         className={`${classNameArrow} ${classNameArrow}${
-          item.getProperties().isExpanded ? '-expanded' : '-collapsed'
+          expandedLayerNames.indexOf(layer.getName()) > -1
```
oterral: USe of `includes`  seems more appropriate

src/components/layertree/LayerTree.js
```diff
@@ -109,106 +136,77 @@ class LayerTree extends Component {
...
-  }
-
-  renderInput(item, level) {
+  renderInput(layer, level) {
     const { classNameInput } = this.props;
     let tabIndex = 0;
 
-    if (!LayerTree.hasChildren(item) || !level) {
-      // We forbid focus on keypress event for first level items and items without children.
+    if (!layer.getChildren().length === 0 || !level) {
```
oterral: No need to compare to 0

oterral:
@mario can you fix the conflict please.

data/TreeData.js
```diff
@@ -1,125 +1,112 @@
...
-          '&style=default&tilematrixset=EPSG%3A3857&Service=WMTS&' +
-          'Request=GetTile&Version=1.0.0&Format=image%2Fpng&',
-        projection: 'EPSG:3857',
+  },
+  sublayers1: {
+    name: 'Others layers',
+    visible: true,
+    type: 'parent',
+    children: {
+      countryBorders: {
```
oterral: Sorry I just see that fxinf the conflict. Why not using an array ? since 'countryBorders' value is never used.

oterral: I mean using an array for the children property.

mhaertwig: I was thinking the same when I saw that the `items` property is an object. I wanted to be consistent but I totally agree when you say we should switch to an array for `items` and `children`.

PR #34, opened 2019-02-27
Remove use of react-i18next

### LINKED ISSUE #21: Remove react-i18next
This lib should translation library agnostic.  We should remove use of react-i18next from ShareMenu and Menu

oterral:
Simplify also Menu and MenuItem components

Fix #21


oterral:
@mhaertwig can you review please ?

PR #39, opened 2019-03-13
Resize the map automatically

oterral:


src/components/ResizeHandler/ResizeHandler.test.js
```diff
@@ -49,37 +50,167 @@ class BasicComponent3 extends React.Component {
...
+            }
+          }}
+        />
+        <ResizeHandler observe={ref} />
+      </>
+    );
+  }
+}
+
+// eslint-disable-next-line  react/prefer-stateless-function
```
edellucien: already disabled globally above no?

PR #44, opened 2019-03-15
Eva/geolocation

gevos:


src/components/Geolocation/Geolocation.scss
```diff
@@ -0,0 +1,80 @@
...
+
+  &.blink {
+    -webkit-animation-name: blinking;
+    -webkit-animation-duration: 3s;
+    -webkit-animation-timing-function: linear;
+    -webkit-animation-iteration-count: infinite;
+    -moz-animation-name: blinking;
+    -moz-animation-duration: 3s;
+    -moz-animation-timing-function: linear;
+    -moz-animation-iteration-count: infinite;
```
oterral: No need of prefixed version of properties. All browser supports it now and if we need it we should use Autoprefixer or something else to generate it automatically.

src/components/Geolocation/Geolocation.scss
```diff
@@ -0,0 +1,80 @@
...
+    animation-iteration-count: infinite;
+  }
+
+  @-moz-keyframes blinking {
+    0% { color: white; }
+    50% { color: $btn-primary-color; }
+    100% { color: white; }
+  }
+
+  @-webkit-keyframes blinking {
```
oterral: same here no need of prefixed version.

src/components/Geolocation/Geolocation.js
```diff
@@ -0,0 +1,213 @@
...
+        feature.changed();
+      }
+    }, 50);
+
+    feature = new Feature({
+      geometry: point,
+      source: new VectorSource(),
+    });
+
+    feature.setStyle(() => {
```
oterral: I would also add a component's property for this `styleFunction` , or at least a `color`  property so we can change the style depending of the app.

src/components/Geolocation/Geolocation.js
```diff
@@ -0,0 +1,237 @@
...
+ * This component displays a geolocator.
+ */
+class Geolocation extends PureComponent {
+  constructor(props) {
+    super(props);
+    const { map, onError, noCenterAfterDrag, colorOrStyleFunc } = this.props;
+
+    this.map = map;
+    this.onError = onError;
+    this.colorOrStyleFunc = colorOrStyleFunc;
```
oterral: Usually we must not set class's property from props's property  otherwise if the props's property is changed by the app  the current will not be aware of this change and will always use the value set from the constructor.
That's why in class's function we always get values from props. 

I hope I'm clear :)

gevos:
Manchmal tritt bei meinen lokalen Tests dieser Fehler auf:

`FAIL  src/components/Geolocation/Geolocation.test.js
  ● Test suite failed to run

    TypeError: Cannot read property 'canvas' of null

      at new CanvasMapRenderer (src/ol/renderer/canvas/Map.js:47:34)
      at Map.createRenderer (src/ol/Map.js:83:22)
      at Map.PluggableMap [as constructor] (src/ol/PluggableMap.js:343:27)
      at new super (src/ol/Map.js:79:5)
      at Object.<anonymous> (src/components/Geolocation/Geolocation.test.js:10:1`

Ich habe noch nicht herausgefunden, wie ich ihn behebe.
@oterral, kam dieser Fehler schon einmal bei dir vor?

oterral:
I look

gevos:
https://github.com/gevos/react-spatial/pull/1 seems to work. Thx

PR #45, opened 2019-03-18
Base layer toggler

oterral:
All in the title.


edellucien:
Very nice MR!
I haven't found anything bothering me here.
@mhaertwig is the assignee so I will let him have a look at it before merging maybe?

mhaertwig:
Very well done!!

PR #48, opened 2019-03-19
Allow setting topiclist padding and for child. With default values

edellucien:


src/components/TopicList/TopicList.js
```diff
@@ -184,7 +195,7 @@ class TopicList extends Component {
           {this.renderInput(topic)}
           {this.renderToggleButton(topic)}
         </div>
-        <div style={{ paddingLeft: `${padding * 2}px` }}>
+        <div style={{ paddingLeft: `${paddingChild}px` }}>
```
oterral: I don't think you need to set padding on the parent  element. If there is a padding to apply to the topicList it will be done at the application level (in TopicListExample`s css in our case).

So knowing that you don't need the `paddingChild` property, use the ` padding`  property directly.

An other thing, we probably want in all case that the `propsToLayerTree.padding` property is equal to the `padding` property.

PR #51, opened 2019-03-20
Implementing copyright

mhaertwig:


src/components/Copyright/Copyright.js
```diff
@@ -0,0 +1,51 @@
...
+
+const propTypes = {
+  layerService: PropTypes.instanceOf(LayerService).isRequired,
+};
+
+class Copyright extends Component {
+  constructor(props) {
+    super(props);
+    const { layerService } = this.props;
+    this.layerService = layerService;
```
oterral: You shouldn't set a class property from a props. You will loose the reference if the props change.

src/components/Copyright/Copyright.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  componentDidMount() {
+    this.updateLayers();
+  }
+
+  componentDidUpdate(prevProps) {
+    const { layerService } = this.props;
+    if (layerService !== prevProps.layerService) {
+      this.layerService.on('change:visible', () => this.updateLayers());
```
oterral: If the layerService object has changed. Here you apply the listener on the wrong object.  You should use the object coming from props.

oterral:
LGTM :v:

PR #52, opened 2019-03-22
Popup: Render popup for small screens (mobile) at the bottom

gevos:


gevos:
> I don't think we have to add a mobile state to the popup . Can you try to use the ResizeHandler component on the parent ?(it will add automatically css class depending on the width and height of the div observed). Then use this class to display the popup at the bottom of its parent when he is too small.

I had thought about this and talked to Mario. He recommended not to use the ResizeHandler but the mobile state. And we still have to handle the mobile behavior in the javascript code as top and left are set in the render function for the desktop view.

oterral:
Top and left can be overridden with !important . 
@mhaertwig  Mobile behavior must be configurable by the application that use react-spatial only with css. I don't think setting an width property for every component that needs a different mobile behavior is healthy to maintain, and what happens when we need a different behvior on small height? another property? totally unmanageable., that's why resizeHandler exists. Moving all this kind logic to css.

oterral:
I'm also thinking about the station plans usecase, when you add the component in an existing  applications. Watching the window size doesn't make sense there.

mhaertwig:
> I had thought about this and talked to Mario. He recommended not to use the ResizeHandler but the mobile state.

I totally agree with @oterral. When I am standing next to the coffee machine in the early morning and you ask me some questions about ways to implement the mobile popup, my answers are neither a "recommendation" nor a review.

src/components/Popup/Popup.md.scss
```diff
@@ -8,3 +8,24 @@
...
+  bottom: 0;
+  top: auto !important;
+  position: fixed;
+  z-index: 2;
+}
+
+.tm-w-xs .tm-popup > div {
+  position: relative;
+  left: 0 !important;
+  bottom: 0 !important;
```
oterral: `!important ` are not need here for left and top
```suggestion
  bottom: 0;
```

src/components/Popup/Popup.md.scss
```diff
@@ -8,3 +8,24 @@
...
+  right: 0;
+  bottom: 0;
+  top: auto !important;
+  position: fixed;
+  z-index: 2;
+}
+
+.tm-w-xs .tm-popup > div {
+  position: relative;
+  left: 0 !important;
```
oterral: `!important ` are not need here for left and top
```suggestion
  left: 0;
```

gevos:
> I don't know if we should display the popup only at the bottom of the PopupExample and not the window.

In the PopupExample the mobile popup is displayed at the bottom of the whole window.

PR #58, opened 2019-04-01
Add basic permalink component

edellucien:


src/Layer.js
```diff
@@ -111,7 +111,7 @@ export default class Layer extends Observable {
   getVisibleChildren() {
     for (let i = 0; i < this.children.length; i += 1) {
       if (this.children[i].getVisible()) {
-        return true;
+        return this.children[i];
```
oterral: it should return an array with all the visible children shouldn't it?

styleguide.config.js
```diff
@@ -3,6 +3,9 @@ const { version } = require('./package.json');
 
 module.exports = {
   title: `React-spatial ${version}`,
+  template: {
+    favicon: 'src/images/favicon.png',
```
oterral: I don't know if it necessary, can you use : path.join(__dirname, ' ?

src/Layer.js
edellucien: Now it does :)

styleguide.config.js
edellucien: I tried with path.join but didn't work atm.

src/Layer.js
edellucien: I mean, now ;)

src/components/Permalink/Permalink.js
```diff
@@ -0,0 +1,68 @@
+import { Component } from 'react';
+import PropTypes from 'prop-types';
+import qs from 'query-string';
+
+const propTypes = {
+  /**
+   * params to be written in url.
```
oterral:
```suggestion
   * Params to be written in url.
```

src/components/Permalink/Permalink.js
```diff
@@ -0,0 +1,68 @@
...
+   */
+  params: PropTypes.object.isRequired,
+
+  /**
+   * either 'react-router' history object:
```
oterral:
```suggestion
   * Either 'react-router' history object:
```

oterral:
Nice one!!

PR #60, opened 2019-04-02
Overwrite styleguidist 'ComponentsList' to allor href to fit with Per…

edellucien:
…malink behavior


edellucien:
Guess it's good now :)

PR #62, opened 2019-04-03
Eva/canvas save button

gevos:


src/components/Checkbox/README.md
```diff
@@ -3,7 +3,7 @@
 This demonstrates the use of Checkbox.
 
 ```jsx
-import React from  'react';
+import React from 'react';;
```
edellucien:
```suggestion
import React from 'react';
```

src/components/TopicList/README.md
```diff
@@ -3,7 +3,7 @@
 This demonstrates the use of TopicList.
 
 ```jsx
-import React from  'react';
+import React from 'react';;
```
edellucien:
```suggestion
import React from 'react';
```

src/components/CanvasSaveButton/CanvasSaveButton.js
```diff
@@ -61,53 +98,138 @@ class CanvasSaveButton extends PureComponent {
...
+        if (extraData && extraData.copyright) {
+          destContext.font = '12px Arial';
+          destContext.fillStyle = 'black';
+          destContext.fillText(extraData.copyright, padding, clip.h - padding);
+        }
+
+        // North arrow
+        if (extraData && extraData.northArrow) {
+          const img = new Image();
+          img.src = extraData.northArrow.circled
```
edellucien: What if we pass a child to NorthArrow, to use it as custom north arrow, such as in this example:
https://github.com/geops/react-spatial/blob/master/src/components/NorthArrow/NorthArrow.test.js#L29
Do you think we should support this scenario here as well?

src/components/CanvasSaveButton/CanvasSaveButton.js
```diff
@@ -61,53 +98,138 @@ class CanvasSaveButton extends PureComponent {
...
+          clip.h,
+        );
+
+        const padding = 5;
+        const arrowSize = 80;
+
+        // Copyright
+        if (extraData && extraData.copyright) {
+          destContext.font = '12px Arial';
+          destContext.fillStyle = 'black';
```
oterral: These properties should also come from the extraData object.

src/components/CanvasSaveButton/README.md
```diff
@@ -0,0 +1,59 @@
...
+    const extent = [-16681616.919511989, 655523.1517989757, -5312679.080488012, 8482674.848201025];
+
+    return (
+      <div className="tm-canvas-save-button-example">
+        <BasicMap
+          map={this.map}
+          center={this.center}
+          zoom={3}
+        />
+        <ShareMenu url={window.location.href}>
```
oterral: I don't think you need the share menu in this example. CanvasSaveButton can work standalone.

src/components/CanvasSaveButton/README.md
```diff
@@ -0,0 +1,59 @@
...
+          <CanvasSaveButton
+            title="Karte als Bild speichern."
+            className='tm-canvas-save-button tm-share-menu-icon'
+            map={this.map}
+            extent={extent}
+            extraData={{
+              copyright: '© Example copyright',
```
oterral: it should be possible to get the copyright from the layerService like this:
```
copyright: layerService.getCopyrights()
```
Does it make sense?

oterral: Can you try to set the copyright dynamically ? if a layers change is visibility the copyright must be updated.

src/components/CanvasSaveButton/README.md
```diff
@@ -0,0 +1,59 @@
...
+              northArrow: {
+                rotation: 25,
```
oterral: Same here the `NorthArrow` component can have a `rotateOffset`, see `NorthArrow` component.
Can you try to set the good rotation of the north arrow  dynamically ?
if the map is rotated the rotation must be updated too.

src/components/CanvasSaveButton/README.md
gevos: I had implemented this, see https://github.com/geops/react-spatial/pull/62/commits/12f9038758ad4a89258bf096b63ced95731ed8b6#diff-84392269f1dcee3ac19b57cd31741243L145. However, after I've talked to @mhaertwig, yesterday, I decided to simplify the properties. So the canvas save button is only responsible for writing on the canvas. Extracting the current copyright and retrieving the map rotation is not handled anymore by the canvas save button.

src/components/CanvasSaveButton/README.md
gevos: See above

src/components/CanvasSaveButton/README.md
mhaertwig: @oterral I think we should keep the component simple. It's good enough to define the copyright manually here. 

However, if we want to reuse the functionality to display copyrights of active layers, we could add a `getActiveLayerCopyrights()` function to the `LayerService` that returns a list of visible layer copyrights.

src/components/CanvasSaveButton/CanvasSaveButton.js
gevos: Done.

src/components/CanvasSaveButton/README.md
gevos: Done.

src/components/CanvasSaveButton/README.md
oterral: That's what I propose in my first comment. I just want to be sure it works , that the component is rerendered when a copyright change. 
Or even better, maybe just manage the case if this copyright property is a function.
```
extraData={{
    copyright: () => {
       return  layerService.getCopyrigths();
   }
}]

src/components/CanvasSaveButton/README.md
oterral: Same answer so :) . We should be able to set a function in the `rotation` property. To calculate the good rotation dynamically.

src/components/CanvasSaveButton/CanvasSaveButton.js
oterral: he's right,`northArrow` to display  should come from the properties of the object as well.

gevos: > Changes I propose :
>     * [...]
>     * Add properties src, width, height to northArrow property to be able to specify a custom northArrow.

Done with src, width, and height

src/components/CanvasSaveButton/README.md
gevos: > Changes I propose :
> 
>     * [...]
> 
>     * Add the possibility to specify a function or a number to the rotation property.

Implemented. Now, you can pass a rotation function or a number

src/components/CanvasSaveButton/README.md
gevos: > Changes I propose :
> 
>     * Add the possibility to specify a function or a number to the copyright property.
Implemented. Now you can pass a copyright function or a text.

I had re-reverted the static getCopyrights() function from the Copyright component. It is used in the CanvasSaveButtonExample. @oterral  suggested in the example above to define a getCopyrights() function in the LayerService. @edellucien @mhaertwig, should I leave it in the Copyright component or should I move it to the LayerService?

gevos:
> Just make an example a bit more dynamic where we can change the visibility of a layer,

Done. By using the getCopyrights() function, only the currently visible copyrights are returned.

src/components/CanvasSaveButton/README.md
edellucien: Das Endergebnis ist die gleiche, aber ich denke auch, es wäre sauberer, es in layerService zu finden. Stimmst du auch dazu @mhaertwig ?

PR #70, opened 2019-04-08
Basic Dialog

edellucien:


src/components/Dialog/Dialog.js
```diff
@@ -0,0 +1,95 @@
...
+      </Header>
+    );
+  }
+
+  render() {
+    const { className, children, isModal, isOpen } = this.props;
+    if (isOpen) {
+      return (
+        <div
+          className={isModal ? 'tm-modal' : null}
```
oterral: the container should always have a className.

src/components/Dialog/Dialog.test.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { configure, mount } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import Dialog from '.';
+
+configure({ adapter: new Adapter() });
+
+describe('Checkbox', () => {
```
oterral:
```suggestion
describe('Dialog', () => {
```

src/components/Dialog/Dialog.js
oterral: I will use `tm-dialog` and `tm-modal`  here and just set a `tm-dialog-content` className on the first child className. 
it's confusing all this modal dialog classes

src/components/Dialog/Dialog.js
```diff
@@ -0,0 +1,95 @@
...
+};
+
+/**
+ * This component creates a Dialog window.
+ */
+class Dialog extends Component {
+  renderDialogTitle() {
+    const { title, onClose, classNameCloseBt } = this.props;
+    return (
+      <Header className="tm-dialog-header">
```
oterral: className should come props

src/components/Dialog/Dialog.js
```diff
@@ -0,0 +1,95 @@
...
+          className={isModal ? 'tm-modal' : null}
+          style={{ display: isOpen ? 'block' : 'none' }}
+        >
+          <div
+            className={`${className} ${
+              isModal ? 'tm-dialog-modal' : 'tm-dialog'
+            }`}
+          >
+            {this.renderDialogTitle()}
+            {children}
```
oterral: Like title, children should be encapsulate in a div.

src/components/Dialog/Dialog.js
```diff
@@ -0,0 +1,138 @@
...
+      isOpen,
+    } = this.props;
+    return (
+      <div
+        className={isModal ? 'tm-modal' : 'tm-dialog'}
+        style={{ display: isOpen ? 'block' : 'none' }}
+      >
+        <div
+          className={`${
+            isModal ? 'tm-modal-container' : 'tm-dialog-container'
```
oterral: you don't need  to differentiate the container className, because the parent already define if it's a modal or not.

edellucien: But it needs a different style however (see scss file)

edellucien:
Changes done.

PR #72, opened 2019-04-09
Create an utils class for KML parsing

oterral:


edellucien:
Clener this way! Thanks

PR #78, opened 2019-04-15
Lucien/dialog props

edellucien:
Allow to pass props to sub-component of Dialog: 'react-draggable'

### DIFF
--- a/src/components/Dialog/Dialog.js
+++ b/src/components/Dialog/Dialog.js
@@ -32,6 +32,12 @@ const propTypes = {
   classNameCloseBt: PropTypes.string,
 
   /**
+   * Object to pass props to sub-component 'react-draggable', to configure it.
+   * https://github.com/mzabriskie/react-draggable
+   */
+  draggableProps: PropTypes.object,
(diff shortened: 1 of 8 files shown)


src/components/Dialog/Dialog.js
```diff
@@ -31,6 +31,12 @@ const propTypes = {
    */
   classNameCloseBt: PropTypes.string,
 
+  /**
+   * Object to pass props to sub-component 'react-draggable', to configure it.
+   * https://github.com/mzabriskie/react-draggable
+   */
+  draggableProps: PropTypes.object,
```
friedjoff: Maybe use Draggable.propTypes instead of PropTypes.object?

edellucien: As discussed I limited the props that we can pass to react-draggable to the minimum, to allow future flexibility of stop using this library.
Therefore, only the 'cancel' props is used, in order to define a selector to be used to prevent drag initialization

friedjoff:
Nice work!

PR #92, opened 2019-05-08
Add documentation how to load SVG differently

oterral:
This fix display of NorthArrow.

And now we use the same image between the NorthArrow component and the CanvasSaveButton.


edellucien:
:heart_eyes:

PR #95, opened 2019-05-09
Allow Text Font (Unsupported in KML) to recover after permalink loadi…

edellucien:
…ng, with ExtendedData


edellucien:
voila :tada:

PR #97, opened 2019-05-10
Create PermalinkField component

edellucien:


src/components/PermalinkField/PermalinkField.js
```diff
@@ -0,0 +1,146 @@
...
+    }
+    document.execCommand('copy');
+  }
+
+  updatePermalinkValue(prevProps) {
+    const { value, getShortenedUrl } = this.props;
+
+    if (value !== prevProps.value) {
+      if (getShortenedUrl) {
+        this.setState({ permalinkValue: getShortenedUrl(value) });
```
oterral: If getShortenedUrl function calls a backend api the result will not be immediat so  setStae must probably be called after the resolve of a promise.

src/components/PermalinkField/PermalinkField.js
```diff
@@ -0,0 +1,146 @@
...
+          value={permalinkValue}
+          readOnly={readOnly}
+          type="text"
+          tabIndex="0"
+          title={titleInputField}
+          className={classNameInputField}
+          ref={node => {
+            this.inputRef = node;
+          }}
+          onChange={() => {}}
```
oterral: eslint complains without the onChange ?

src/components/PermalinkField/PermalinkField.js
```diff
@@ -0,0 +1,146 @@
...
+  titleCopyBt: '',
+  titleInputField: '',
+  button: <MdContentCopy focusable={false} />,
+  getShortenedUrl: null,
+};
+
+/**
+ * This component displays a permalink field.
+ */
+class PermalinkField extends PureComponent {
```
oterral: I will call it a PermalinkInput since we already have a SearchInput

src/components/PermalinkField/PermalinkField.js
```diff
@@ -0,0 +1,146 @@
...
+      titleInputField,
+    } = this.props;
+
+    const { permalinkValue } = this.state;
+
+    return (
+      <div className={className}>
+        <input
+          value={permalinkValue}
+          readOnly={readOnly}
```
oterral: readonly will always be true, you can remove it from the props.

src/components/PermalinkField/PermalinkField.md.scss
```diff
@@ -0,0 +1,27 @@
...
+    input {
+      padding: 0 0 0 15px;
+    }
+
+    div[role='button'] {
+      border-left: 1px solid $gray;
+      width: 30px;
+      display: flex;
+      align-items: center;
+      justify-content: center;
```
oterral: this button's css should go in PermalinkField.scss.   Have a look to SearchInput default css. The css are pretty similar.

styleguide.config.js
```diff
@@ -91,6 +91,7 @@ module.exports = {
             'src/components/List/[A-Z]*.js',
             'src/components/ListItem/[A-Z]*.js',
             'src/components/Permalink/[A-Z]*.js',
+            'src/components/PermalinkField/[A-Z]*.js',
```
oterral: Should be in Form section since it's an input.

src/components/PermalinkField/PermalinkField.js
edellucien: Yes not readonly input without onChange creates warning message in the console. But we don't always want readonly, even if we don't react on change

src/components/PermalinkField/PermalinkField.js
edellucien: The wkp permalinkInput from drawing tool isn't readOnly. I would assume that readOnly isn't default param.

edellucien: But you are right, let's put readOnly as default, and we can still make it configurable if needed later.

edellucien:
Changes according to your comments

edellucien:
Ready for another review

src/components/PermalinkInput/PermalinkInput.js
```diff
@@ -0,0 +1,151 @@
...
+
+const defaultProps = {
+  value: '',
+  className: 'tm-permalink-field',
+  classNameInputField: 'tm-permalink-input',
+  classNameCopyBt: 'tm-permalink-bt',
+  titleCopyBt: '',
+  titleInputField: '',
+  button: <MdContentCopy focusable={false} />,
+  getShortenedUrl: null,
```
oterral: set a dflt function that returns a promise . With this default value you can get rid of the `if` lin71

edellucien: Done :)

edellucien:
Tests added

src/components/PermalinkInput/PermalinkInput.js
```diff
@@ -0,0 +1,151 @@
...
+
+  constructor(props) {
+    super(props);
+    this.state = { permalinkValue: null };
+    this.inputRef = null;
+  }
+
+  componentDidMount() {
+    const { value, getShortenedUrl } = this.props;
+    getShortenedUrl(value).then(v => {
```
oterral:
```suggestion
    getShortenedUrl(value).then(permalinkValue => {
```

src/components/PermalinkInput/PermalinkInput.js
```diff
@@ -0,0 +1,151 @@
...
+      this.setState({ permalinkValue: v });
```
oterral:
```suggestion
      this.setState({ permalinkValue });
```

src/components/PermalinkInput/PermalinkInput.js
```diff
@@ -0,0 +1,151 @@
...
+    if (this.inputRef) {
+      this.inputRef.select();
+    }
+    document.execCommand('copy');
+  }
+
+  updatePermalinkValue(prevProps) {
+    const { value, getShortenedUrl } = this.props;
+
+    if (value !== prevProps.value) {
```
oterral: Move this if in component componentDidUpdate so you can also use the updatePermalinkValue function in componentDidMount

src/components/PermalinkInput/README.md
```diff
@@ -0,0 +1,26 @@
...
+This demonstrates the use of PermalinkInput.
+
+```jsx
+import React from 'react';
+import PermalinkInput from 'react-spatial/components/PermalinkInput';
+
+class PermalinkInputExample extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { value: window.location.href };
```
oterral: Non need to use this.state since you don't update it

src/components/PermalinkInput/README.md
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  render() {
+    const { value } = this.state;
+    return (
+      <div className="tm-permalink-input-example">
+        <PermalinkInput value={value} />
```
oterral:
```suggestion
        <PermalinkInput value={window.location.href} />
```

package.json
```diff
@@ -57,6 +57,8 @@
     "ole": "^0.0.3"
   },
   "dependencies": {
+    "babel-polyfill": "^6.26.0",
+    "flush-promises": "^1.0.2",
```
oterral: Do we really need these 2 modules ?

edellucien: Aimed to flush promises to test Promise more easily in jest.

PR #103, opened 2019-05-15
Pass react-draggable 'onStop' callback directly, with all parameters

edellucien:


edellucien:
onStop by default returns two params: evt and position.

friedjoff:
👍

PR #104, opened 2019-05-16
Don't use default image for CanvasSaveButton

oterral:
I think it  makes it more easier to setup CSS for button in the application.


edellucien:
Netlify deploy fails however @oterral

oterral:
you can rereview @edellucien

src/components/CanvasSaveButton/CanvasSaveButton.js
```diff
@@ -131,11 +141,34 @@ class CanvasSaveButton extends PureComponent {
...
+
         if (extent) {
-          const pixelTopLeft = map.getPixelFromCoordinate(getTopLeft(extent));
-          const pixelBottomRight = map.getPixelFromCoordinate(
-            getBottomRight(extent),
-          );
+          firstCoordinate = getTopLeft(extent);
+          oppositeCoordinate = getBottomRight(extent);
+        } else if (coordinates) {
+          // In case of coordinates coming from DragBox interaction:
```
gevos: This explanation would be nice at the prop description

gevos:
Except for the small docu comment, LGTM

src/components/CanvasSaveButton/CanvasSaveButton.js
oterral: It's comment for developers. A user does not care, he will send an array of coordinates and expects it works.

PR #107, opened 2019-05-17
Eva/kml background fill color

gevos:


src/components/FeatureExportButton/FeatureExportButton.test.js
```diff
@@ -144,6 +144,9 @@ describe('FeatureExportButton', () => {
 
       const textStyle = new Style({
         text: new Text({
+          backgroundFill: new Fill({
+            color: 'rgba(255,255,255,0.01)',
+          }),
```
oterral: backgroundFill is added to `<ExtendedData>` tag in the kml file.  This test only tests `<Style>`  tag . Can you add a test for ExtendedData please ?

gevos: see https://github.com/geops/react-spatial/pull/108

PR #112, opened 2019-05-22
[WIP] TEst with hidpi layer

oterral:


oterral:
I close this since we already export hidpi layer

PR #113, opened 2019-05-22
added requestEncoding option to wmts config

danji90:


src/ConfigReader.js
```diff
@@ -50,6 +50,7 @@ class ConfigReader {
         zIndex: -1,
         source: new WMTSSource({
           url: item.data.url,
+          requestEncoding: item.data.requestEncoding,
```
oterral: @danji90 can you add a test in config reader pls.

PR #117, opened 2019-05-24
[TGSRV-121] Fix hack for ExtendedData...

gevos:
...Move each ExtendedData section to the correct position.

### DIFF
--- a/src/utils/KML.js
+++ b/src/utils/KML.js
@@ -271,23 +271,28 @@ const writeFeatures = (layer, featureProjection) => {
 
     // Temporary to locate <ExtendedData> after <Style>
     // TODO: Remove it when https://github.com/openlayers/openlayers/pull/9500 is merged and ol updated.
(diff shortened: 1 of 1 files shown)


oterral:
Can you give an example of what was wrong with the previous code  please. it will help to review .

gevos:
> Can you give an example of what was wrong with the previous code please. it will help to review .

You can find an example in the ticket, referenced in the commit message. Draw 3 lines in plan-editor and generate a kml by trying to save them.

oterral:
yes but it's a react-spatial bug so if we can have a test case with react-spatial, it's better.  So everybody can test wihtout installing plan editor.

edellucien:
I tested it in plan-editor and works to create correct KML for multiple lines. Good catch and thanks for fixing Eva! We will have to revert this temporary solution when ol will publish the fix :)

oterral:
thx all

PR #121, opened 2019-05-29
Add permalink prop 'coordinateDecimals'

edellucien:


src/components/Permalink/Permalink.js
```diff
@@ -99,12 +105,17 @@ class Permalink extends Component {
...
-      y: center[1],
+      x: this.roundCoord(center[0]),
+      y: this.roundCoord(center[1]),
       z: mapView.getZoom(),
     });
   }
 
+  roundCoord(val) {
+    const { coordinateDecimals } = this.props;
+    return val.toFixed(coordinateDecimals).replace(/[.,]00$/, '');
```
oterral:
```suggestion
    return parseFloat(val.toFixed(coordinateDecimals));
```
` parseFloat` removes the 00 and 0 at the end automatically

edellucien: Thanks :)

PR #125, opened 2019-06-03
Unable default drag/pinch rotate on map interactions (Add this intera…

edellucien:
…ction for north arrow example).

### DIFF
--- a/src/components/BasicMap/BasicMap.js
+++ b/src/components/BasicMap/BasicMap.js
@@ -1,7 +1,10 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { defaults as defaultInteractions } from 'ol/interaction';
-import { DragRotate, PinchRotate } from 'ol/interaction';
+import {
+  defaults as defaultInteractions,
+  DragRotate,
+  PinchRotate,
+} from 'ol/interaction';
 
 import OLMap from 'ol/Map';
 import OLCollection from 'ol/Collection';
--- a/src/components/BasicMap/BasicMap.js
+++ b/src/components/BasicMap/BasicMap.js
@@ -136,24 +136,6 @@ class BasicMap extends Component {
       zoom,
     } = this.props;
 
-    if (map) {
-      if (interactions) {
-        interactions.forEach(i => {
-          map.addInteraction(i);
-        });
-      } else {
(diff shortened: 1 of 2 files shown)


src/components/BasicMap/BasicMap.js
```diff
@@ -132,11 +136,34 @@ class BasicMap extends Component {
       zoom,
     } = this.props;
 
+    this.dfltInteractions = defaultInteractions({
```
oterral:
```suggestion
    const dfltInteractions = defaultInteractions({
```

src/components/BasicMap/BasicMap.js
```diff
@@ -132,11 +136,34 @@ class BasicMap extends Component {
...
+        map
+          .getInteractions()
+          .getArray()
+          .forEach(i => {
+            if (i instanceof DragRotate || i instanceof PinchRotate) {
+              map.removeInteraction(i);
+            }
+          });
+      }
+    }
```
oterral: I don't think you need this block . If user provide a Map with interactions it's up to him to manage his interactions.

src/components/BasicMap/BasicMap.js
```diff
@@ -132,11 +136,34 @@ class BasicMap extends Component {
...
+
     this.map =
       map ||
       new OLMap({
         controls: [],
-        interactions: interactions || defaultInteractions(),
+        interactions: interactions || this.dfltInteractions,
```
oterral: Put the defaultinteractiond efinition here:

```
efaultInteractions({
      altShiftDragRotate: false,
      pinchRotate: false,
    });
```

src/components/BasicMap/BasicMap.js
edellucien: My requirement for this task is to deactivate rotation,pinch rotate, etc by default in react-spatial.

edellucien: This requirement will not be fulfilled, if I don;t unset by default rotation. But still allow the user to add interactions directly in props if needed.

edellucien: quote from mario: 'Wenn keine Map übergeben wird, initialisieren wir die Karte ohne Rotation.'
so I removed it :+1:

PR #127, opened 2019-06-03
Responsive Sidebar added

angsherpa232:
This is ResponsiveSideNavbar that is responsive to different screen sizes. Users can enter Sidebar Item. If this work sample is OK then the further process will be to include the icon option in sidebar item and toggling to iconOnly option as well.


oterral:
Thx for this !! but I close it since we already merge the other sidebar.

PR #155, opened 2019-06-24
updated Button and sidebar item, added accesskey as prop

danji90:
+ updated snapshots

### DIFF
--- a/src/components/Button/Button.js
+++ b/src/components/Button/Button.js
@@ -28,6 +28,11 @@ const propTypes = {
   title: PropTypes.string,
 
   /**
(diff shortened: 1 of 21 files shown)


src/components/Button/Button.js
```diff
@@ -39,14 +44,23 @@ const defaultProps = {
   title: undefined,
   tabIndex: 0,
   style: undefined,
+  accessKey: '',
```
oterral: undefined by default so the attribute is not added when empty
```suggestion
  accessKey: undefined,
```

PR #156, opened 2019-06-25
Fix KML sanitize function on read

edellucien:


src/components/FeatureStyler/FeatureStyler.js
```diff
@@ -277,61 +277,77 @@ class FeatureStyler extends PureComponent {
...
+                text: textStyle,
+                image: iconStyle,
+                zIndex: oldStyle.getZIndex(),
+              }),
+            );
+          });
+        }
+      }
+      return resolve(
+        new Style({
```
oterral: I guess, you can create a style variable before  the  previous if and use it to avoid duplication of Style declaration

src/components/FeatureStyler/FeatureStyler.js
```diff
@@ -277,61 +277,77 @@ class FeatureStyler extends PureComponent {
...
+                fill: fillStyle,
+                stroke: strokeStyle,
+                text: textStyle,
+                image: iconStyle,
+                zIndex: oldStyle.getZIndex(),
+              }),
+            );
+          });
+        }
+      }
```
oterral: use the `else` here

src/components/FeatureStyler/FeatureStyler.js
```diff
@@ -529,12 +545,14 @@ class FeatureStyler extends PureComponent {
       textRotation,
     });
 
-    // Set feature's properties
-    feature.set('name', text);
-    feature.set('description', description);
+    stylePromise.then(newStyle => {
+      // Set feature's properties
+      feature.set('name', text);
+      feature.set('description', description);
```
oterral: No need to d this inside the promise

PR #157, opened 2019-06-25
Use hidpi wmts

oterral:
Just change the scale property to 1,2 or 3  in the CanvaSaveButon example to test it.


oterral:
@edellucien  can you test  ad bit please but don't merge it I need to set the example as it was before

edellucien:
Tested sucessfully, MR looks good!

PR #161, opened 2019-06-27
Don't save in KML text features with empty text value

edellucien:


src/components/FeatureExportButton/FeatureExportButton.test.js
```diff
@@ -136,7 +137,7 @@ describe('FeatureExportButton', () => {
       delete unnamedlayer.name;
       const wrapper = mount(<FeatureExportButton layer={unnamedlayer} />);
       const exportString = wrapper.instance().createFeatureString(unnamedlayer);
-      expect(/<name>/g.test(exportString)).toBe(false);
+      expect(/<name>ExportLayer<\/name>/g.test(exportString)).toBe(false);
```
oterral: Why the exportString can't be tested now ?

edellucien: because KML creates <name></name> for layer name and Placemark name. So I wanted to test more specifically the layer name with this regex (ExportLayer).

src/utils/KML.js
```diff
@@ -237,7 +237,14 @@ const writeFeatures = (layer, featureProjection) => {
 
     const olStyle = new Style(newStyle);
     clone.setStyle(olStyle);
-    exportFeatures.push(clone);
+
+    // Don't save empty text feature
+    if (
+      !clone.getStyle().getText() ||
+      (clone.getStyle().getText() && clone.get('name').length)
```
oterral: wrong if you only test if there is a getText(). 
You should test if :
```
if (clone.geometry() instanceof Point && !style.getImage() && style.getText() && !style.getText().getText()) {
  // don't export;
}
```

src/components/FeatureExportButton/FeatureExportButton.test.js
oterral: so test `<document><name>ExportLayer `   becuase <name> could be a placemark's  name

src/utils/KML.js
```diff
@@ -237,7 +237,17 @@ const writeFeatures = (layer, featureProjection) => {
 
     const olStyle = new Style(newStyle);
     clone.setStyle(olStyle);
-    exportFeatures.push(clone);
+
+    if (
+      clone.getGeometry() instanceof Point &&
+      olStyle.getText() &&
+      !olStyle.getText().getText().length
```
oterral:
```suggestion
      !olStyle.getText().getText()
```
no need to use length

src/utils/KML.js
```diff
@@ -237,7 +237,17 @@ const writeFeatures = (layer, featureProjection) => {
...
+    ) {
+      // Don't save empty text feature
+      exportFeatures.push(new Feature());
```
oterral: Here you save an empty feature. We just don'want to export features at all

edellucien: If there is one empty text, but several other features, we still want to export them, but without the empty one.
exportFeatures is an array of features to save in the kml

edellucien: Do ensure KML.writeFeatures() not to fail, I need to pass at least an empty feature.

src/utils/KML.js
edellucien: self commited

src/utils/KML.js
oterral:
```
if (
     !( clone.getGeometry() instanceof Point &&
      olStyle.getText() &&
      !olStyle.getText().getText())
    ) {
      exportFeatures.push(clone);
    }
```

edellucien: That must be the final commit :angel:

PR #164, opened 2019-06-28
WIP: Use hooks for permalinkInput component

edellucien:
Use hooks in permalinkInput component

TODO: find a way to test functional component with hooks


oterral:
@edellucien is it finished?

PR #165, opened 2019-06-28
Remove 'ReadOnly', because it doesn't allow to select in ie, therefor…

edellucien:
…e not to copy on button click


oterral:
Doesn't seem to work

oterral:
Sorry it works I thought you fixed the copy button

PR #184, opened 2019-07-22
Allow to change line colors and closure icons

edellucien:
- Style line closure icon selector
- allow config of colors for stroke


mhaertwig:
Please let's not use the `tm-` namespace for react spatial. This namespace was for another application and we should clean this up to avoid collisions.

mhaertwig:
I merge this and create a ticket for the cleaning up the namespaces.

PR #185, opened 2019-07-25
Hide baseLayerToggler if only one baselayer is present

edellucien:


edellucien:
Can be tested by using 'exampleData' instead of 'treeData'

mhaertwig:
Nice, thanks.

PR #187, opened 2019-07-29
Lucien/layer doc

edellucien:


edellucien:
Better squash that together :+1:

PR #195, opened 2019-08-08
Upgrade to ol6 and add Mapbox layer

oterral:
This PR contains the update of openlayers a a new Mapbox layer to display vector tiles.

Closes #196


src/layers/MapboxLayer.js
```diff
@@ -0,0 +1,79 @@
...
+import Layer from '../Layer';
+
+/**
+ * A class representing vector layer to display on BasicMap
+ * @class
+ */
+export default class MapboxLayer extends Layer {
+  constructor(options = {}) {
+    super(options);
+    this.styleUrl = options.url;
```
mhaertwig: @oterral let's use arrow functions.

oterral: @mhaertwig What do you mean ?

oterral: done

PR #197, opened 2019-08-13
Eva/hide in legend

gevos:


src/components/LayerTree/LayerTree.js
```diff
@@ -205,6 +205,10 @@ class LayerTree extends Component {
       ? []
       : [...layer.getChildren()];
 
+    if (layer.getHideInLegend()) {
```
mhaertwig: @gevos you could do this check on top of the function. this increases readability and a little bit the performance (e.g. the `includes` check is not necessary if you return afterwards).

PR #198, opened 2019-08-13
WIP: Adds layer method getFeatureInfoAtCoordinate()

mhaertwig:
@oterral as disucssed: what's your general opinion?
If you agree with the code, I will add some test. So far, I didn't test the function for the WMSLayer.

### DIFF
--- a/src/Layer.js
+++ b/src/Layer.js
@@ -142,7 +142,6 @@ export default class Layer extends Observable {
    * @param {ol.Coordinate} coordinate Coordinate to request the information at.
    * @returns {Promise<Object>} Promise with features, layer and coordinate
    *  or null if no feature was hit.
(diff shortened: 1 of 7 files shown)


src/Layer.js
```diff
@@ -130,4 +130,26 @@ export default class Layer extends Observable {
...
+    // This layer has no onClick.
+    // The function is implemented by inheriting layers.
+  }
+
+  /**
+   * Request feature information for a given coordinate.
+   * @param {ol.Coordinate} coordinate Coordinate to request the information at.
+   * @returns {Promise<Object>} Promise with features, layer and coordinate
+   *  or null if no feature was hit.
+   * eslint-disable-next-line class-methods-use-this
```
oterral: you don't need this anymore , since you use this in the function
```suggestion
```

PR #214, opened 2019-08-27
Create 'isItemExpanded' prop to handle default toggle in layerTree

edellucien:


mhaertwig:
This should happen automatically and without additional parameter. A layer group is expanded, if the layer has visible children (see internal ticket).

oterral:
@mhaertwig good idea, and easier to implement thx.

edellucien:
Indeed, new MR : https://github.com/geops/react-spatial/pull/215

PR #215, opened 2019-08-28
Toggle layer with no visible children

edellucien:


src/components/LayerTree/LayerTree.js
```diff
@@ -93,9 +93,19 @@ const defaultProps = {
 class LayerTree extends Component {
   constructor(props) {
     super(props);
+    const { layerService, isItemHidden } = this.props;
+    const initialExpandedLayerNames =
```
mhaertwig: Couldn't we use the layer instances itself instead of their keys (`expandedLayers`)?

edellucien: Both are working. I do it with the layer instance instead.

src/components/LayerTree/LayerTree.js
```diff
@@ -93,9 +93,18 @@ const defaultProps = {
...
+      layerService && layerService.getLayers()
+        ? layerService
+            .getLayers()
+            .filter(l => !isItemHidden(l))
+            .filter(l => l.getVisibleChildren().length)
```
oterral: one filter to rules them all
```
.filter(l => !isItemHidden(l) &&l.getVisibleChildren().length )
```

edellucien: better so!

edellucien:
Ready to be merged

PR #220, opened 2019-08-29
Terminate MapboxLayer when layer is removed...

gevos:
...Otherwise new mapbox container elements would be created over and over again.


src/layers/MapboxLayer.js
```diff
@@ -68,6 +68,12 @@ export default class MapboxLayer extends Layer {
       return;
     }
 
+    this.map.getLayers().on('remove', evt => {
+      if (evt.element === this.olLayer) {
+        this.terminate();
+      }
+    });
```
oterral: unlisten this event in terminat function

PR #222, opened 2019-08-31
Eva/base layer toggler toggle fallback

gevos:


src/components/BaseLayerToggler/BaseLayerToggler.js
```diff
@@ -152,6 +170,23 @@ class BaseLayerToggler extends Component {
         this.map.getView().setCenter(coord);
       }
     });
+
+    map.on('moveend', () => {
```
oterral: Always remove event in componentWillUnmount

PR #230, opened 2019-09-10
SelectColors Component 

angsherpa232:


src/components/SelectColors/SelectColors.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+const SelectColors = function(props) {
+  const { options } = props;
+  const { className } = props;
+  const { onChange } = props;
+  if (!options) {
+    return null;
+  }
+
+  return React.createElement(
```
oterral: We use https://fr.reactjs.org/docs/introducing-jsx.html

src/components/SelectColors/SelectColors.js
```diff
@@ -0,0 +1,74 @@
...
+
+const defaultProps = {
+  options: undefined,
+  className: 'tm-selectcolors',
+};
+
+/**
+ * This component displays a simple selectColors HTML element.
+ */
+const SelectColors = function(props) {
```
edellucien: Can import props directly here, with functional component:
const SelectColors = ({ options, className, onChange }) => {

src/components/SelectColors/SelectColors.js
edellucien: Nepalese version of the doc is not yet ready, here is the english one :)
https://en.reactjs.org/docs/introducing-jsx.html

edellucien: You can check other functional components example in the lib, and follow the same pattern, such as:
https://github.com/geops/react-spatial/blob/master/src/components/Select/Select.js

edellucien:
For each component, we create a README.ms file, that is used by styleguidist, to have a live example.
For the user, but as well for us, it is easier to test your component.
such as: https://github.com/geops/react-spatial/blob/master/src/components/Select/README.md

src/components/SelectColors/SelectColors.js
angsherpa232: Done

src/components/SelectColors/SelectColors.js
angsherpa232: I guess I referred to the older one where React.createElement was used. I updated as jsx in this PR.

edellucien: Yes I commented at the same time you commited the update.

edellucien: But it is still relevant to import directly the function delcaration:
'''
const SelectColors = ({ options, className, onChange }) => {
''''
instead of 
'''
const SelectColors = props => {
  const { options, className, onChange } = props;
'''

angsherpa232: Done, thanks for the suggestion.

edellucien:
We need to be able to see the currently selected color in the select

edellucien:
Additionally, would be better to use a functional component example, such as:
https://deploy-preview-230--react-spatial.netlify.com/#checkbox
It is more readable

src/components/SelectColors/README.md
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  render() {
+    return (
+      <div className="tm-selectcolors-example">
+        <SelectColors
+          options={this.colors}
+          value={this.state.color}
+          onChange={color => {
+            let index = this.colors.findIndex(x => x.name === color);
```
edellucien: can use const here

edellucien:
We decided to use a react color library instead. I will close this MR.
Thanks for the work anyway Ang! :+1:

PR #233, opened 2019-09-11
Ensure evt are well unregistered from layerService

edellucien:


src/components/Copyright/Copyright.js
```diff
@@ -45,12 +45,16 @@ class Copyright extends Component {
...
-    layerService.unlistenChangeEvt('change:visible');
+    if (layerService) {
+      layerService.un('change:visible', this.onChangeRef);
+    }
   }
 
   updateLayerService() {
     const { layerService } = this.props;
-    layerService.on('change:visible', () => this.updateCopyright());
+    this.onChangeRef = layerService.on('change:visible', () =>
```
oterral: this.onChangeRef is undefined I guees `on` does not return the callback.

src/components/LayerTree/LayerTree.js
```diff
@@ -151,7 +158,9 @@ class LayerTree extends Component {
     const { layerService } = this.props;
     if (layerService) {
       this.updateLayers(layerService);
-      layerService.on('change:visible', () => this.updateLayers(layerService));
+      this.onChangeRef = layerService.on('change:visible', () =>
```
oterral: With this method you have to `un` before the  `on`  because the internale `un` will not work since you create another callback.

PR #245, opened 2019-09-17
Allow to force resize of mapBox layer

edellucien:


edellucien:
My plan was to register the on change:size in the target app, but it is even better here, you are right!

PR #315, opened 2019-11-05
Update popup components with props comments and a new classNameWrappe…

edellucien:
…r prop


edellucien:
Mr updated after your comments.

mhaertwig:
Please make the title bold like in the screen design.

PR #326, opened 2019-11-06
Fix for panning popups into view.

mhaertwig:
@oterral the functionality for panning the popup into the map's viewport is finished now. Maybe you can have a short look on the whole functionality. I pushed this on master last Sunday :-/

### DIFF
--- a/src/components/Popup/Popup.js
+++ b/src/components/Popup/Popup.js
@@ -136,7 +136,7 @@ class Popup extends PureComponent {
     if (
       panIntoView &&
       popupElement &&
-      (popupElement !== prevState.popupElement || feature !== prevProps.feature)
+      popupElement !== prevState.popupElement
     ) {
       this.panIntoView();
     }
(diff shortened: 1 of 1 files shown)


src/components/Popup/Popup.js
```diff
@@ -234,6 +234,10 @@ class Popup extends PureComponent {
 
     const { top, left } = this.state;
 
+    // force re-render if the feature or the coordinate changes.
+    // this is needed to update the popupElement ref
+    const key = feature ? feature.getId() : popupCoordinate.join();
```
oterral: Did you try to use createRef to get the element. It will do the work without this.

src/components/Popup/Popup.js
```diff
@@ -244,6 +248,7 @@ class Popup extends PureComponent {
       >
         <div
           role="presentation"
+          key={key}
           ref={popupElement => {
             this.setState({ popupElement });
```
oterral: doing this is on non-sense, use createRef instead

mhaertwig: The problem is that I need to measure the dimensions of the popup _after_ the popup was rendered. I use the key for forcing a re-render if the coordinate or the feature was updated. Consequently the state is updated and I trigger another component update.

Maybe there's a better solution but I didn't find one so far.

src/components/Popup/Popup.js
mhaertwig: Should work as well. However, I need the key to force an update of the component.

PR #357, opened 2019-11-21
Cleaning up the base layer toggler.

mhaertwig:


src/components/BaseLayerToggler/BaseLayerToggler.js
```diff
@@ -330,43 +284,55 @@ class BaseLayerToggler extends Component {
...
-        </Footer>
+          </div>
+        </div>
       );
     }
 
     const nextLayer = layers[idx];
 
     return (
       <div className={className} ref={this.ref}>
```
oterral: Since className is a common html attribute. I've put the default value here (see other cleanup) and then we apply the other props .
```
const { titles, ...other } = this.props;
      <div className='rs-base-layer.toggler'  ref={this.ref} {...other} >
```
What do you think?

friedjoff: The advantage of keeping it as a default prop would be, to have it appear in the documentation. If you just want to modify the default style with a custom className a little bit, you could add the default classname together with a custom className without looking into the code.

PR #359, opened 2019-11-21
Cleaning up the Geolocation control.

mhaertwig:


mhaertwig:
Ahaha.
The button needs to be removed though.

src/components/Geolocation/Geolocation.js
```diff
@@ -60,8 +53,7 @@ const propTypes = {
 };
 
 const defaultProps = {
-  className: 'tm-geolocation',
-  classNameActive: 'blink',
+  className: 'rs-geolocation',
   title: undefined,
```
oterral: title and className can be remove using spread props int eh render method.

friedjoff: To support the `active` class I kept the className prop. Also this improves documentation so we should think about doing this always.

src/components/Geolocation/Geolocation.js
```diff
@@ -220,17 +206,25 @@ class Geolocation extends PureComponent {
...
     const { active } = this.state;
 
     return (
-      <Button
-        className={`${className} ${active ? classNameActive : ''}`}
-        title={title}
+      <div
+        role="button"
+        tabIndex="0"
+        className={`${className} ${active ? 'rs-blink' : ''}`}
```
oterral: @friedjoff  maybe or more generic class like ` rs-active` is more suitable.

src/components/Geolocation/Geolocation.js
oterral: MAybe  you 're rigth. Can you add the classNme  for all the other components? thx

friedjoff:
done

PR #367, opened 2019-11-22
Cleanup Popup

oterral:


src/components/Popup/Popup.js
```diff
@@ -46,64 +50,92 @@ const propTypes = {
...
    */
-  title: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  onCloseClick: PropTypes.func,
 
   /**
-   * Popup padding.
+   * Hide or show the header.
    */
-  padding: PropTypes.string,
+  showHeader: PropTypes.bool,
```
friedjoff: How about replacing this prop by allowing the user to set `renderHeader: () => null`? Same would apply to `showCloseButton`.

src/components/Popup/Popup.js
```diff
@@ -46,64 +50,92 @@ const propTypes = {
...
-  onKeyUp: () => {},
+  titles: { closeButton: 'Close' },
   onCloseClick: () => {},
-  t: p => p,
+  renderHeader: null,
+  renderCloseButton: null,
+  renderFooter: () => null,
 };
 
 class Popup extends PureComponent {
```
friedjoff: Why not a functional component?

src/components/Popup/Popup.js
```diff
@@ -197,74 +229,64 @@ class Popup extends PureComponent {
...
     const {
       feature,
       showHeader,
       popupCoordinate,
-      className,
       children,
-      onKeyUp,
-      padding,
-      tabIndex,
+      title,
```
friedjoff: Maybe rename `title` to header to avoid confusion with `titles`?

src/components/Popup/Popup.js
oterral: Yes i've thought the same .

src/components/Popup/Popup.js
oterral: I've tried but there was a problem with the refresh and the panIntoView option. @mhaertwig fixed it in the version so I choose to let it like that since I couldn't succeed to have a better version.

src/components/Popup/Popup.js
oterral: yaes

src/components/Popup/Popup.js
oterral: Done

src/components/Popup/Popup.js
oterral: Done

src/components/Popup/README.md
```diff
@@ -97,6 +100,7 @@ class PopupExample extends React.Component {
           title="Geometry"
```
friedjoff:
```suggestion
          header="Geometry"
```

PR #369, opened 2019-11-22
Cleanup Zoom

oterral:


src/components/Zoom/Zoom.js
```diff
@@ -53,112 +35,101 @@ const propTypes = {
...
+    };
+  }, [map, zoomSlider, ref.current]);
+
+  return (
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    <div className="rs-zooms-bar" {...other}>
+      <div
+        role="button"
+        tabIndex={0}
+        className="rs-zoom-in rs-round-blue"
```
friedjoff: Shouldn't the className be neutral to how it looks? Why do we need `rs-round-blue`?

src/components/Zoom/Zoom.js
```diff
@@ -53,112 +35,101 @@ const propTypes = {
...
+      }
+    },
+    [map],
+  );
+
+  useEffect(() => {
+    let control;
+    if (zoomSlider && ref.current) {
+      control = new ZoomSlider();
+      control.element.classList.remove('ol-control');
```
friedjoff: Why do we need this? A comment would be nice? Also the next line is unclear to me.

src/components/Zoom/Zoom.js
oterral: agree. I can move the style  to the example css file instead.

src/components/Zoom/Zoom.js
oterral: To void the apply of some default ol css I guess. @edellucien do you remember why?

oterral: I remove this line. But leave the enxt one it's to avoid TAB navigation on this button.

src/components/Zoom/Zoom.js
oterral: Done

src/components/Button/Button.scss
```diff
@@ -8,23 +8,23 @@
   transition: background-color 0.5s ease, color 0.5s ease;
 }
 
-.tm-round-blue {
+.rs-round-blue {
```
friedjoff: This class can be removed now, right?

oterral: I guess yess

PR #384, opened 2019-12-03
Update libs

oterral:


friedjoff:
url-loader seems to be incompatible with NodeJS 8 which is end-of-life by the end of the year. Do you agree with removing NodeJS 8 from CI and add version 12 instead?

oterral:
@friedjoff yes . I thought it was already the case.

oterral:
@friedjoff  fixed

PR #386, opened 2019-12-03
Export radToDeg method from NorthArrow to fix canvasSaveButton example

edellucien:


friedjoff:
How about moving `radToDeg` to utils? I can imagine other places where this function might be useful. The current implementation makes it pretty hidden and easy to brake in the future.

oterral:
Maybe use a existing very small npm module for this. so we don't have to care about this anymore.

src/components/NorthArrow/NorthArrow.js
```diff
@@ -71,4 +70,6 @@ function NorthArrow({ map, rotationOffset, circled, children, ...other }) {
 NorthArrow.propTypes = propTypes;
 NorthArrow.defaultProps = defaultProps;
 
-export default React.memo(NorthArrow);
+const memoized = React.memo(NorthArrow);
+
+export default memoized;
```
oterral: go back to the oald version

oterral:
@edellucien rebase fix then yiou can merge.

PR #387, opened 2019-12-04
Add test workflow for github

edellucien:


friedjoff:
Can you remove the travis config file and add Yarn caching support (see https://github.com/geops/tree-app/blob/master/.github/workflows/cypress.yml for an example)?

oterral:
`Successful in 1m` that's a gooooooood CI.

PR #419, opened 2019-12-19
Remove non spatial

friedjoff:


src/components/LayerTree/LayerTree.js
```diff
@@ -240,26 +221,29 @@ class LayerTree extends Component {
...
-        }`}
-        className={classNameToggle}
-        onClick={() => {
-          this.onInputClick(layer, layer.getChildren().length);
-        }}
+      <div
+        role="button"
+        tabIndex={0}
+        className="rs-layer-tree-toggle"
+        title={title}
```
edellucien: Could we add here 'aria-expanded' for accessibility:
https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/expanded/

edellucien:
```suggestion
        title={title}
        aria-expanded={expandedLayerNames.includes(layer)}
```

friedjoff:
The broken Netlify build has been fixed, but the preview seems still broken. Anybody any idea?

friedjoff:
Should work now. Problem was solved by switching all CSS class names to `rs-` prefix.

friedjoff:
> The LayerTree does not work anymore: The checkboxes and radio button do not react.

Finally fixed by removing one line.

friedjoff:
@oterral as discussed I've removed the TopicList component.

PR #423, opened 2019-12-20
Gevos/behig improvements

edellucien:


src/components/BaseLayerToggler/BaseLayerToggler.js
```diff
@@ -97,19 +107,30 @@ class BaseLayerToggler extends Component {
...
       if (!layers.length) {
         return;
       }
-      const children = layers[idx].getChildren();
-      const childLayers = children.length ? children : [layers[idx]];
+      let children = [];
+      let childLayers = [];
+      if (idx !== null) {
+        children = layers[idx].getChildren();
+        childLayers = children.length ? children : [layers[idx]];
```
oterral: could create a [undefined] array if layers is empty

src/components/BaseLayerToggler/BaseLayerToggler.js
```diff
@@ -129,7 +150,19 @@ class BaseLayerToggler extends Component {
   }
 
   componentWillUnmount() {
+    const { layerService } = this.props;
     unByKey([this.postRenderKey, this.moveEndKey]);
+    layerService.un('change:visible', e => this.updateState(e));
```
oterral: you unlisten a different function , it should be the same that you use when you listen.

src/components/BaseLayerToggler/BaseLayerToggler.js
```diff
@@ -138,26 +171,61 @@ class BaseLayerToggler extends Component {
     if (!layerService) {
       return;
     }
-
+    layerService.un('change:visible', e => this.updateState(e));
+    layerService.un('change:layers', () => this.resetState());
     this.updateState();
-    layerService.on('change:visible', () => this.updateState());
+    layerService.on('change:visible', e => this.updateState(e));
+    layerService.on('change:layers', () => this.resetState());
```
oterral: Same here

oterral: Same here

src/components/BaseLayerToggler/BaseLayerToggler.js
```diff
@@ -138,26 +171,61 @@ class BaseLayerToggler extends Component {
...
+    if (!layerVisible) {
+      return;
+    }
+
+    const layers = layerService.getBaseLayers() || [];
+    const idx = layers.indexOf(layerVisible);
+    let newIdx;
+
+    if (idx === -1) {
+      newIdx = 1;
```
oterral: Why 1 an d not 0 ?

src/components/BaseLayerToggler/BaseLayerToggler.js
edellucien: But I listen/unlisten the same:
`('change:visible', e => this.updateState(e));`
Doesn't it work so?

src/components/BaseLayerToggler/BaseLayerToggler.js
edellucien: Because if the previous visible doesn't exist for the new topic, the first baselayer will be set as visible for the map, and it the next one that needs to be used for the BaseLayerToggler.

src/components/BaseLayerToggler/BaseLayerToggler.js
edellucien: The is a return if !layers.length just above, so it should be safe isn't it?

oterral: good point :+1:

oterral: but `layers[idx]` could be null ?

src/components/BaseLayerToggler/BaseLayerToggler.js
oterral: you create 2 differents objects . How the js engine is suppose to know the are the same function.

src/components/BaseLayerToggler/BaseLayerToggler.js
edellucien: I added a condition to prevent this scenario

src/components/BaseLayerToggler/BaseLayerToggler.js
oterral: I believe you on this one ;)

PR #429, opened 2020-01-07
Filter undefined copyright from Copyrights component.

edellucien:


src/components/Copyright/Copyright.js
```diff
@@ -35,7 +35,8 @@ function Copyright({ layerService, format, ...other }) {
         layerService
           .getLayersAsFlatArray()
           .filter(l => l.getVisible())
-          .map(l => l.getCopyright()),
+          .map(l => l.getCopyright())
+          .filter(cr => cr !== undefined),
```
gevos:
```suggestion
          .filter(cr => cr),
```
The copyright can be `"",` e.g. when it comes from Mapbox source and `sourceCache.used` is `false` see https://github.com/geops/react-spatial/pull/418

gevos:
@edellucien Please, also have a look at https://github.com/geops/trafimage-maps/pull/408

src/components/Copyright/Copyright.js
edellucien: So you want to filer on '' and false as well?

gevos: @edellucien Yes, I want to filter out `""`. And in which case is it `false`?

PR #482, opened 2020-02-07
Mouse position: Add initial index

gevos:


gevos:
> Since the MousePosition should work like a select box, it will be nicer if the component have a 'value' property where you put the current projection object to be selected.
> 
> ```
> const projs = [proj1, proj2];
> conse selected = projs[1];
> <MousePosition value={proj1} projections={projs} onChange={(proj)=> {
>   selected = proj;
> }} />
> ```

Done

friedjoff:
I would prefer something like `projectionValue` instead of `value`, because without looking at the documentation I don't know what value means.

gevos:
> I would prefer something like `projectionValue` instead of `value`, because without looking at the documentation I don't know what value means.

@oterral suggested `value`, so it looks like select box. But I agree, `projectionValue` sounds better. I have renamed it.

PR #483, opened 2020-02-07
Map context

friedjoff:
Example implentation for map context.

Things to discuss:

- Where do we put the children DOM elements in BasicMap?
- Do we want to create directories for different component types, like layers and controls?

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,6 +121,13 @@
           "devDependencies": true
         }
       ],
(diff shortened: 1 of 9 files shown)


lb1c:
Ping @friedjoff @mhaertwig

friedjoff:
I was not able to convince the co-maintainers.

PR #528, opened 2020-03-18
BaseLayerSwitcher component

danji90:
BaseLayerToggler is to be replaced by the BaseLayerSwitcher to boost performance and usability
Implemented:
* Functional component structure
* Main functionality (open/close menu button, change layer on layer click)
* Load Static images for buttons (including option of loading from layer when not provided in props)
* Tests
* Open/Close CSS transition


src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,161 @@
...
+          title={titles.closeSwitcher}
+        >
+          <FaChevronLeft focusable={false} />
+        </div>
+      ) : null;
+
+    return (
+      <div
+        className={`${className}${switcherOpen ? ' open' : ''}`}
+        ref={this.ref}
```
mhaertwig: Avoid using `refs` if they are not really necessary!

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,161 @@
...
+    if (map) {
+      this.setState({ layers });
+    }
+  }
+
+  render() {
+    const { className, titles } = this.props;
+    const {
+      layers,
+      fallbackImg,
```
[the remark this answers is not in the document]
danji90: Removed

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,161 @@
...
+  className: 'rs-base-layer-switcher',
+  // fallbackImgDir: '../../images/baselayer/',
+  // validExtent: [-Infinity, -Infinity, Infinity, Infinity],
+  titles: {
+    button: 'Open Baselayer-Switcher',
+    closeSwitcher: 'Close Baselayer-Switcher',
+  },
+};
+
+class BaseLayerSwitcher extends Component {
```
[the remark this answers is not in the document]
danji90: Done

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+
+  render() {
+    const { className, titles } = this.props;
+    const {
+      layers,
+      fallbackImg,
+      fallbackImgOpacity,
+      switcherOpen,
```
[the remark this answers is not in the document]
danji90: Done

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,161 @@
...
+import OLMap from 'ol/Map';
+import LayerService from '../../LayerService';
+
+import './BaseLayerSwitcher.scss';
+
+const propTypes = {
+  /**
+   * An ol.Map.
+   */
+  map: PropTypes.instanceOf(OLMap).isRequired,
```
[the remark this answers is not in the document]
danji90: Removed, importing layers as props

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,161 @@
...
+
+  /**
+   * CSS class to apply on the container.
+   */
+  className: PropTypes.string,
+
+  /**
+   * Path to the directory which includes the fallback images
+   */
+  // fallbackImgDir: PropTypes.string,
```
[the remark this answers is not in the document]
danji90: Done

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: Removed, along with all Class Component stuff

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,168 @@
...
+    /* Used for correct layer image render with animation */
+    if (!switcherOpen) {
+      window.clearTimeout(timeout);
+      timeout = window.setTimeout(() => {
+        setIsClosed(true);
+      }, 200);
+      return;
+    }
+    setIsClosed(false);
+  }, [switcherOpen]);
```
oterral: this useEffect should return a function that clears also  the timeout

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,168 @@
...
+        <FaChevronCircleLeft size={15} focusable={false} />
+      </div>
+    ) : null;
+
+  return (
+    <div className={`${className}${switcherOpen ? ' open' : ''}`}>
+      <div className="rs-base-layer-switch-layers">
+        {switcherOpen ? toggleBtn : null}
+        {baseLayers.map((layer, index) => {
+          const layerName = layers[index].getName();
```
oterral:
```suggestion
          const layerName = layer.getName();
```

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,168 @@
...
+          return (
+            <div
+              key={baseLayers[index].key}
```
oterral: you already have the layer in 'layer' prop

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,168 @@
...
+              className={`rs-base-layer-switch-button${
+                switcherOpen ? ' layer' : ''
```
oterral: add 'rs-' prefix

oterral: I guess you wanna call this class 'open' or 'close'

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,168 @@
...
+              }${
+                baseLayers[index].getName() === currentLayer.getName()
+                  ? ' active'
```
oterral: prefix

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,178 @@
...
+};
+
+const getNextImage = (currentLayer, layers, layerImages) => {
+  const currentIndex = layers.indexOf(
+    layers.find(layer => layer === currentLayer),
+  );
+  const nextIndex = currentIndex + 1 === layers.length ? 0 : currentIndex + 1;
+  return layerImages.filter(layer => layer)[
+    Object.keys(layerImages)[nextIndex]
+  ];
```
oterral: Is `layerImages[nextIndex]` not enough ?

danji90: I love overcomplicating things XD Thanks

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: did you mean like this?

oterral: No see the doc https://de.reactjs.org/docs/hooks-effect.html#effects-with-cleanup

danji90: Second try:
`useEffect(() => {
    /* Used for correct layer image render with animation */
    if (!switcherOpen) {
      const timeout = setTimeout(() => {
        setIsClosed(true);
      }, 200);
      return () => clearTimeout(timeout);
    }
    return setIsClosed(false);
  }, [switcherOpen]);`

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,181 @@
...
+   */
+  layerImages: PropTypes.objectOf(PropTypes.string),
+
+  /**
+   * CSS class to apply on the container.
+   */
+  className: PropTypes.string,
+
+  /**
+   * Alternaive text rendered if layer images can't be loaded
```
oterral:
```suggestion
   * Alternative text rendered if layer images can't be loaded
```

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,181 @@
...
+              >
+                {isClosed ? titles.button : layerName}
+              </div>
+              <img
+                src={
+                  isClosed
+                    ? getNextImage(currentLayer, baseLayers, images)
+                    : images[index]
+                }
+                alt={altText}
```
oterral:
```suggestion
                alt={t(altText)}
```

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
```diff
@@ -0,0 +1,74 @@
...
+      border: 2px solid white;
+      background-color: gray;
+
+      &.open {
+        &:hover {
+          border: 2px solid red;
+        }
+
+        &.rs-active {
+          border: 2px solid red;
```
oterral: duplicated css, move it with &:hover

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
```diff
@@ -0,0 +1,74 @@
...
+    .rs-base-layer-switch-button {
+      position: relative;
+      height: 85%;
+      min-width: 100px;
+      margin: 4px;
+      overflow: hidden;
+      border: 2px solid white;
+      background-color: gray;
+
+      &.open {
```
oterral: i think this one is useless. hover and active css can be applied without it.

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
```diff
@@ -0,0 +1,74 @@
...
+        }
+      }
+
+      .rs-base-layer-switch-image {
+        vertical-align: center;
+        width: 100px;
+        height: 100px;
+      }
+
+      img[alt] {
```
oterral: is there image without alt ?

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
```diff
@@ -0,0 +1,74 @@
...
+
+  .rs-base-layer-switch-layers {
+    display: flex;
+    overflow: auto;
+    height: 100%;
+    width: 360px;
+
+    .rs-base-layer-switch-button {
+      position: relative;
+      height: 85%;
```
oterral: why 85% ?

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
```diff
@@ -0,0 +1,74 @@
...
+      min-width: 100px;
```
oterral: i think putting a fixed with make more sense

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    .rs-base-layer-switch-close {
+      display: flex;
+      flex-direction: column;
+      justify-content: center;
+      height: 85%;
+      margin: 4px;
+      background-color: white;
+      border: 2px solid white;
```
oterral: this class a common css with the other buttons, move them at the same place.

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
```diff
@@ -0,0 +1,74 @@
...
+  height: 80px;
+  width: 115px;
+  transition: 250ms width;
+  overflow: hidden;
+
+  &.open {
+    width: 360px;
+  }
+
+  .rs-base-layer-switch-layers {
```
oterral: Don't put child classes to this class if it's not necessary. it wil be easier to overide if we need it .
Forexamples:  
```
.rs-base-layer-switcher { ...
   .rs-base-layer-switch-layers {...}
   .rs-base-layer-switch-button {...}
   .rs-base-layer-switch-title {...}
}

File: src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss (hunk 16 above)
Comment (danji90) [reply src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss]: I only want the border effects when the switcher is open, if I remove it the border will be red also when closed

File: src/components/BaseLayerSwitcher/BaseLayerSwitcher.js (hunk 7 above)
Comment (oterral) [reply src/components/BaseLayerSwitcher/BaseLayerSwitcher.js]: You were close :
```
useEffect(() => {
    /* Used for correct layer image render with animation */
   let timeout;
    if (!switcherOpen) {
      timeout = setTimeout(() => {
        setIsClosed(true);
      }, 200);
      } else {
        setIsClosed(false);
     }
      return () => clearTimeout(timeout);
    }
    return setIsClosed(false);
    return () => clearTimeout(timeout);
  }, [switcherOpen]);

```

File: src/components/BaseLayerSwitcher/BaseLayerSwitcher.js [hunk 22]
```diff
@@ -0,0 +1,181 @@
...
+          const layerName = layer.getName();
+          return (
+            <div
+              key={layer.key}
+              className={`rs-base-layer-switch-button${
+                switcherOpen ? ' open' : ''
+              }${layerName === currentLayer.getName() ? ' rs-active' : ''}`}
+              role="button"
+              title={isClosed ? titles.openSwitcher : layerName}
+              aria-label={isClosed ? titles.openSwitcher : layerName}
```
oterral: Duplicate code

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,181 @@
...
+              <img
+                src={
+                  isClosed
+                    ? getNextImage(currentLayer, baseLayers, images)
+                    : images[index]
+                }
+                alt={altText}
+                className="rs-base-layer-switch-image"
+              />
+            </div>
```
oterral: Maybe toi be more understandable, create to <div>
```
isClosed ?<><div  className={`rs-base-layer-switch-title rs-closed}>titles.button</div><img ...></>
: <><div  className={`rs-base-layer-switch-title}>layerName</div><img ...></>
```

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,181 @@
...
+                  if (!switcherOpen) {
+                    setSwitcherOpen(true);
+                  } else {
+                    layer.setVisible(true);
+                    setCurrentLayer(layer);
+                    setSwitcherOpen(false);
+                  }
+                }
+              }}
+              tabIndex={`${isClosed ? index : '0'}`}
```
oterral: I don't think  we need different tabIndex , put 0 everywhere. it could potentially breaks the default keyboard navigation.

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,181 @@
...
+              }}
+              onKeyPress={e => {
+                if (e.which === 13) {
+                  if (!switcherOpen) {
+                    setSwitcherOpen(true);
+                  } else {
+                    layer.setVisible(true);
+                    setCurrentLayer(layer);
+                    setSwitcherOpen(false);
+                  }
```
oterral: onKeyPress and onClick are very similar.

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,181 @@
...
+      <div className="rs-base-layer-switch-layers">
+        {switcherOpen ? toggleBtn : null}
+        {baseLayers.map((layer, index) => {
+          const layerName = layer.getName();
+          return (
+            <div
+              key={layer.key}
+              className={`rs-base-layer-switch-button${
+                switcherOpen ? ' open' : ''
+              }${layerName === currentLayer.getName() ? ' rs-active' : ''}`}
```
oterral: MAkes it more understandable :  

const openClass = switcherOpen ? '  rs-open' : ';
const activeClass =  '{layerName === currentLayer.getName() ? ' rs-active' : ''}`}

<div className={`rs-base-layer-switch-button${openClass}${activeClass} />

oterral: same for the closedClass `rs-closed`

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,181 @@
...
+        title={titles.closeSwitcher}
+      >
+        <FaChevronCircleLeft size={15} focusable={false} />
+      </div>
+    ) : null;
+
+  return (
+    <div className={`${className}${switcherOpen ? ' open' : ''}`}>
+      <div className="rs-base-layer-switch-layers">
+        {switcherOpen ? toggleBtn : null}
```
oterral:
```suggestion
        {switcherOpen && toggleBtn}
```

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: If I give all of them "0" it will skip to the next button even when the switcher is closed. Ideal would be giving the first button 0 and the others -1 while closed so they are omitted when using tab.

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: Thanks

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
danji90: Removed

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
danji90: I only left the min-width

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
danji90: done

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
danji90: Yes that makes a lot of sense

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: moved the code to variable

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: More lines of code, but fair enough

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: Storing the code in function and calling when needed

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: Yes this makes sense, thanks for pointing it out

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: I wrote a function that returns the tabIndex we need

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: Done

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
danji90: I removed it

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
oterral: ok

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,198 @@
...
+  const setTabIndex = (closed, buttonIndex) => {
+    if (closed) {
+      return buttonIndex === 0 ? '0' : '-1';
+    }
+    return '0';
+  };
+
+  if (!baseLayers || baseLayers.length < 2) {
+    return null;
+  }
```
oterral: I think this 'if' must be moved after all the useEffect function , because hook must  always be executed in the same order. It's surprising eslint is not complaining.

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,198 @@
...
+        setIsClosed(true);
+      }, 200);
+    } else {
+      setIsClosed(false);
+    }
+    return () => clearTimeout(timeout);
+  }, [switcherOpen]);
+
+  const toggleBtn =
+    baseLayers.length > 0 ? (
```
oterral: At this point baseLayers.length  is always > 0 so you can remove the if

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,198 @@
...
+          const activeClass =
+            layerName === currentLayer.getName() ? ' rs-active' : '';
+          return (
+            <div
+              key={layer.key}
+              className={`rs-base-layer-switcher-button rs-layer${openClass}${activeClass}`}
+              role="button"
+              title={buttonTitle}
+              aria-label={buttonTitle}
+              onClick={() => onButtonSelect(switcherOpen, layer)}
```
oterral: no need to pass switcherOpen in the parameter of the function.

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,198 @@
...
+              onKeyPress={e => {
+                if (e.which === 13) {
+                  onButtonSelect(switcherOpen, layer);
+                }
+              }}
+              tabIndex={setTabIndex(isClosed, index)}
```
oterral: same here no need to pass isClose in the function.

and it's `getTabIndex` not  `setTabIndex`

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,198 @@
...
+        aria-label={titles.closeSwitcher}
+        title={titles.closeSwitcher}
+      >
+        <FaChevronCircleLeft size={15} focusable={false} />
+      </div>
+    ) : null;
+
+  return (
+    <div className={`${className}${openClass}`}>
+      <div className="rs-base-layer-switcher-layers">
```
oterral: Do we really need this element ?

src/components/BaseLayerSwitcher/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+const center = [1149722.7037660484, 6618091.313553318];
+const map = new OLMap({ controls: [] });
+const layers = ConfigReader.readConfig(treeData);
+const layerService = new LayerService(layers);
+const layerImages = {
+  osm: osmImage,
+  osmhot: osmhotImage,
+  opentopo: openTopoImage,
+};
```
oterral: Where is travic used  ?

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
oterral: > If I give all of them "0" it will skip to the next button even when the switcher is closed (to buttons that aren't visible).

that means after the user goes trough all the element with tabIndex 0, it will goes to element with tabIndex 1 then 2 and so on. So it just pushes the problem later. The question you must ask yourself , why the focus goes to hidden elements ?  maybe the hidden elements should not be there.

if the menu is closed displays only the button that opens the menu, if it's open, displays all the base layers button and the button to close it.

danji90: The reason I don't just render all the buttons on switcherOpen was because I couldn't get it to work with the transition animation. Will try to find a way though

danji90: You're right, in the meantime I can use the isClosed variable instead of switcherOpen so the animation runs smoothly on open and close

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: Thanks for pointing it out

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: Removed this function in the meantime, don't need it anymore

src/components/BaseLayerSwitcher/README.md
danji90: Some residual stuff from when I was testing with geOps tile layers. Removing

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: Nope, removing it

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: true, I can call it directly

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,196 @@
...
+              aria-label={layerName}
+              onClick={() => onLayerSelect(layer)}
+              onKeyPress={e => {
+                if (e.which === 13) {
+                  onLayerSelect(layer);
+                }
+              }}
+              tabIndex="0"
+            >
+              <>
```
oterral: No need of this anymore

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,196 @@
...
+                  alt={altText}
+                  className="rs-base-layer-switcher-image"
+                />
+              </>
+            </div>
+          );
+        })
+      ) : (
+        <div
+          key={0}
```
oterral: useless

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  .rs-base-layer-switcher-title {
+    position: absolute;
+    width: 100%;
+    font-size: 10px;
+    padding: 2px 0;
+    background-color: rgba(255, 255, 255, 0.9);
+    text-align: center;
+    z-index: 2;
```
oterral: Can you make it work without the z-index ?

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
```diff
@@ -0,0 +1,58 @@
...
+
+    &.rs-closed {
+      color: white;
+      background-color: rgba(0, 0, 0, 0.7);
+    }
+  }
+
+  .rs-base-layer-switcher-image {
+    width: 100px;
+    height: 100px;
```
oterral: why the height is 100px ? it goes outside the parnet div.

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
```diff
@@ -0,0 +1,58 @@
...
+  display: flex;
+
+  &.rs-open {
+    width: 600px;
+  }
+
+  .rs-base-layer-switcher-button {
+    margin: 4px;
+    background-color: white;
+    border: 2px solid white;
```
oterral: maybe using display flex here, will fix the issue i mentionned further.

oterral: forget I've seen there is an opacity on  the title.

danji90: This has been fixed using flex, since the image is inserted as background now

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
danji90: img has been removed

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
danji90: also removed due to flexbox

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: agreed

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,199 @@
...
+                  onLayerSelect(layer);
+                }
+              }}
+              style={getImageStyle(images[index])}
+              tabIndex="0"
+            >
+              <div className="rs-base-layer-switcher-title">{layerName}</div>
+              {images[index] ? null : (
+                <span className="rs-alt-text">{altText}</span>
+              )}
```
oterral: Is this a requirement this altText ?

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
```diff
@@ -0,0 +1,63 @@
...
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+  }
+
+  .rs-layer {
+    min-width: 100px;
+    background-color: rgb(187, 187, 187);
+  }
```
oterral: This class is useless.  I still believe you should put a fixed width to the rs-base-layer-switcher-button class, otherwise if the text  is too long the div width will different between layers button.

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: @mhaertwig did mention I should put in props

src/components/BaseLayerSwitcher/BaseLayerSwitcher.scss
danji90: OK, but I need to pass in a min-width (equal to the width), otherwise the buttons will be squashed during the animation

danji90: And also I will have go specify width (and min-width) for the close button as well

danji90: Or I suppose I could just give that button a different class

danji90: @oterral done, I removed the class and gave removed some more unnecessary css

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -0,0 +1,199 @@
...
+            </div>
+          );
+        })
+      ) : (
+        <div
+          className="rs-base-layer-switcher-button rs-layer"
```
oterral:
```suggestion
          className="rs-base-layer-switcher-button"
```

PR #532, opened 2020-03-20
[LayerTree] hide items with hideInLegend property by default

candux:
The hideInLegend property is used in the demos and example data. The need for a isItemHidden function is not really documented so it's a hassle to figure out why it's not working out of the box.
I guess hiding also the baselayers by default would be to invasive


oterral:
`hideInLegend` is just an example how to use `isItemHidden` , that's why it's used only in example. It has nothing to do with the core code

PR #533, opened 2020-03-21
[Geolocation] add alwaysRecenterToPosition property

candux:
I find the current Geolocation component unintuitive. By default,
it will recenter to the current position every second on my smartphone.

If noCenterAfterDrag is set, it will
- center to the current position if the map hasn't been dragged yet
- don't center to the position if the map has been dragged. Even when
  you disable the component and enable it again.

Services like Google Maps center once to the current position and
then every time you press the the location-button again.
I think most users expect such behavior.

I therefor introduced a new property alwaysRecenterToPosition.
By default or if noCenterAfterDrag is set, the behavior stays the
same.
If alwaysRecenterToPosition is set to false, the map will recenter to
the current position and every time the component is enabled again.

Also rename isCentered to recenterToPostion


oterral:
I guess the google behavior should be the default one.

candux:
Making it the only option would make the logic easier. But you would have to change the property name at the same time in trafimage

src/components/Geolocation/Geolocation.js
```diff
@@ -66,10 +72,10 @@ class Geolocation extends PureComponent {
       source: new VectorSource(),
     });
 
-    this.isCentered = true;
+    this.recenterToPostion = true;
```
oterral, line 75: Can we rename this property to `isRecenteringToPosition`  otherwise we could think it's a function.

oterral:
let's merge #548 before this one

src/components/Geolocation/Geolocation.js
candux: sure

oterral:
@candux conflict !!!!!

candux:
conflict fixed

oterral:
@candux thx

PR #537, opened 2020-03-25
LayerTree: Add isAlwaysExpanded option...

gevos:
# Description

...If a layer (in the layer tree) is expanded,
all its children are also expanded if they have the option isAlwaysExpanded=true

This works recursively.

# Others

- [x] IE11 tested.
\- These changes are done for the topic zweitausbildung in trafimage-maps which does not have a release, yet.
- [x] Tests added.


src/components/LayerTree/LayerTree.js
```diff
@@ -101,13 +101,15 @@ class LayerTree extends Component {
...
-            )
+        ? this.getExpandedLayers(
+            layerService
+              .getLayers()
+              .filter(
+                l =>
+                  !isItemHidden(l) &&
+                  l.getVisibleChildren().filter(c => !isItemHidden(c)).length,
+              ),
+          )
```
oterral: Is  ` this.getExpandedLayers(layerService.getLayers())`  not enough ?

src/components/LayerTree/LayerTree.js
```diff
@@ -149,12 +151,31 @@ class LayerTree extends Component {
...
+   * Get the always expanded ancestors (isAlwaysExpanded=true) of the given layers
+   * together with the (given) initially expanded layers
+   *
+   * @param  {Layer} layers Initially expanded layers
+   * @return {Array.<Layer>} Initially expanded layers and all its always expanded ancestors
+   */
+  getExpandedLayers(layers) {
+    const { isItemHidden } = this.props;
+    const children = layers.flatMap(l =>
+      l.getChildren().filter(c => !isItemHidden(c) && c.getIsAlwaysExpanded()),
```
oterral: In addition to comment line 112   should n't test  here also if `!isItemHidden(l) && l.getIsAlwaysExpanded() === true`  ? 

I don't know if it makes sense I just want to avoid to have the same filter in 2 different places.

src/components/LayerTree/LayerTree.js
gevos: In the first level we expand all layers which are not hidden and have visible children that are not hidden.
Beginning with the second level, we expand the layers that are not hidden and have the attribute `isAlwaysExpanded`, no matter if its children are visible or not. If one of these layers is expanded, we look at the next level (recursively).
So, the rules are different for the first level and the following levels.

src/components/LayerTree/LayerTree.js
gevos: No, if we keep the behavior of https://github.com/geops/react-spatial/pull/537#discussion_r398411227 as the first level is treated differently. So, the given layers are always expanded in this function.

gevos:
>  I think you can remove also the toggle functionnality from the display if the layer is always expanded.

Done

PR #538, opened 2020-03-26
Update libs

oterral:


candux:
After updating to mapbox-gl 1.9.0 I can't see any OL-Layers anymore. Do you see similiar behavior? If not, I can try to make a testcase.

oterral:
We haven't enough tested , IE 11 is also broken :(

oterral:
@candux  Probably this is the problem https://github.com/geops/react-spatial/pull/553/files

candux:
> @candux Probably this is the problem https://github.com/geops/react-spatial/pull/553/files

Great. I will check it out maybe this evening

PR #539, opened 2020-03-26
Allow to change geolocation children

edellucien:


edellucien:
> I don't see the test . You probably forgot to `git add` it ;)
> 
> otherwise it's perfect thx.

Exactely I just forgot to add it! :stuck_out_tongue_closed_eyes:

edellucien:
Here it is :tada:

PR #540, opened 2020-03-26
Allow to change prev/next baselayer button's children

edellucien:


src/components/BaseLayerToggler/BaseLayerToggler.js
```diff
@@ -24,6 +24,16 @@ const propTypes = {
    */
   className: PropTypes.string,
 
+  /**
+   * Children of the previous baselayer button.
+   */
+  previousLayerBtChildren: PropTypes.node,
```
friedjoff: How about naming these props `prevButtonContent` and `nextButtonContent`? This would match the `titles.prevButton` prop naming in line 52, 53 and does not leak too much implementation details.

edellucien: Yes good proposition thanks!

PR #542, opened 2020-03-27
reduce CPU usage  of Geolocation Component

candux:
I noticed a very high CPU usage of the geolocation component. Even after I disabled all animations, I still see about 30% CPU usage with no updates and 60% with 1 update/s.
Reasons are
* the layer, style and feature are recreated for every location Update
* even if you provide your own style function, the timer for the animation is still created.

We can instead reuse the layer, style and just update the location.
With this patch I'm down to about 0% CPU usage with no update and 10% at 1 update/s.

With animations enabled you will not see so much improvments.

### DIFF
--- a/src/components/Geolocation/Geolocation.js
+++ b/src/components/Geolocation/Geolocation.js
@@ -75,6 +75,7 @@ class Geolocation extends PureComponent {
 
     this.state = {
       active: false,
+      point: undefined,
     };
   }
 
@@ -116,6 +117,7 @@ class Geolocation extends PureComponent {
   }
 
   activate(position) {
+    const { point } = this.state;
     const { map } = this.props;
 
     const code = map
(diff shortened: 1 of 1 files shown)


friedjoff:
Nice catch, thanks a lot!

src/components/Geolocation/Geolocation.js
```diff
@@ -127,10 +130,17 @@ class Geolocation extends PureComponent {
...
+    if (!point) {
+      var tmpPoint = new Point(pos);
+      this.setState({
+        point: tmpPoint,
+      });
+      this.highlight(tmpPoint);
+      this.layer.setMap(map);
+    } else {
+      point.setCoordinates(pos);
+    }
```
oterral: 12 lines of code instead of 3  only to avoid the creation of a point once in a while, it seems a bit overkill. Optimisation is good only when you gain something (performance or code readibility).   And  putting the point in the state is totally useless  since you never update it, so you trigger a useless render of the component.

oterral: I see where the problem was . I thought activate was the function call when you click on the button which is not . So this make sense but the setState is still useless.

candux: Yes, the function gets called for every location update which can be very often. This change makes a huge performace difference on my smartphone

> but the setState is still useless.
I need to store the point somewhere to update the position. Can I do that without the setState?

candux: I will try to refactor it that way

PR #543, opened 2020-03-29
[mapbox layer] don't rerender if unnecessary => make everything faster

candux:
I noticed that as soon as I have a mapbox-layer enabled, many openlayers
operations like updating a style, moving a feature or zooming becomes
slower and the fan of my computer is constantly running. This is also the
case even if the operation has noting to do with the mapbox layer.
Without a mapbox-layer everything is super fast.

As it turns out, openlayers is calling render() on every layer if anything
changes anywhere. At the moment, this always triggers a complete rendering
in mapbox-gl even if nothing changed there. On my PC, this wastes about
10ms. If there are many OL-Layers this can happen very often.

We only have to render the mapbox-gl layer again, if something within it
changed.
This is also what others do:
https://github.com/uber/react-map-gl/blob/master/src/mapbox/mapbox.js

Fix this situation by keeping track of the most recent render run and
only rerender if something changed.

This helps operations which change OL-Layers a lot. Updating a style 
every 100ms did use about 40% CPU. Now it's nearly nothing.

Also the initial loading, zooming etc, becomes faster, if there
a many layers.


src/layers/MapboxLayer.js
```diff
@@ -33,27 +33,42 @@ export default class MapboxLayer extends Layer {
   constructor(options = {}) {
     const mbLayer = new OLLayer({
       render: frameState => {
+        var changed = false;
         const canvas = this.mbMap.getCanvas();
         const { viewState } = frameState;
 
         const visible = mbLayer.getVisible();
-        canvas.style.display = visible ? 'block' : 'none';
+        if (this.renderState.visiable != visible) {
```
oterral:
```suggestion
        if (this.renderState.visible != visible) {
```

src/layers/MapboxLayer.js
```diff
@@ -33,27 +33,42 @@ export default class MapboxLayer extends Layer {
...
+          canvas.style.display = visible ? 'block' : 'none';
+          this.renderState.visiable = visible;
```
oterral:
```suggestion
          this.renderState.visible = visible;
```

oterral:
Nice one!!! Thx!!!

src/layers/MapboxLayer.js
```diff
@@ -33,27 +33,46 @@ export default class MapboxLayer extends Layer {
   constructor(options = {}) {
     const mbLayer = new OLLayer({
       render: (frameState) => {
+        let changed = false;
         const canvas = this.mbMap.getCanvas();
         const { viewState } = frameState;
 
         const visible = mbLayer.getVisible();
-        canvas.style.display = visible ? 'block' : 'none';
+        if (this.renderState.visiable !== visible) {
```
oterral, line 41:
```suggestion
        if (this.renderState.visible !== visible) {
```

PR #544, opened 2020-03-29
[geolocation] make it possible to pass a real function as style

candux:
Motivation:
I use a directional icon for the location and rotate it by the device's
compass. After a compass change, I update the style.
Afterwards you have to call feature.changed() so the change gets processed.

For this reason I need to prove a real function to the geolocation
component, which receives the feature as a parameter and returns a style.

This patch also handles the case that someone passes a plain style as it
was possible before.

### DIFF
--- a/src/components/Geolocation/Geolocation.js
+++ b/src/components/Geolocation/Geolocation.js
@@ -203,8 +203,10 @@ class Geolocation extends PureComponent {
           circleStyle,
         ];
       });
-    } else {
+    } else if (colorOrStyleFunc instanceof Style) {
       feature.setStyle(colorOrStyleFunc);
+    } else {
+      feature.setStyle(colorOrStyleFunc(feature));
     }
 
     this.layer.getSource().clear();


friedjoff:
Sounds like a good idea! We should add a test and documentation for this feature. @candux do you want to do this? Otherwise I can do it.

candux:
@friedjoff I will give it a shot.

@oterral Not anymore, see https://github.com/geops/react-spatial/pull/542 . I think a static timer is also just not good enough.

candux:
@friedjoff I didn't manage to create a meaningfull test. Can you please do that? Thanks!

oterral:
@candux are you able to make a piece of code to show the problem on codepen.io or sandbox.io?

candux:
> @candux are you able to make a piece of code to show the problem on codepen.io or sandbox.io?

Heres the code I use to rotate the compass:
```javascript
   geolocationStyleFunction(feature) {
        const { styleLocation } = this.state;
        this.setState({
            geolocationFeature: feature
        });
        return styleLocation;
    }

    handleOrientation = (event) => {
        const { styleLocation,  geolocationFeature } = this.state;
        if (geolocationFeature) {
            var compassdir;
             if(event.webkitCompassHeading) {
              // Apple works only with this, alpha doesn't work
              compassdir = event.webkitCompassHeading;  
            }
            else {
                compassdir = 360 - event.alpha;
            }
            styleLocation.image_.rotation_ = (compassdir) * (2* Math.PI / 360);
            geolocationFeature.changed();
        }
    }

    componentDidMount() {
        // TODO: only when needed
        window.addEventListener("deviceorientationabsolute", this.handleOrientation);
    }

 <Geolocation 
                map = { this.map }
                alwaysRecenterToPosition={false}
                colorOrStyleFunc={this.geolocationStyleFunction}
            />
```
This PR is needed so I can execute `geolocationFeature.changed();`

I can try to build a working example of it but it will maybe take a few days until I have time.

oterral:
That's what I thought you can already do this with the current status
``` 
const { styleLocation, feature } = this.state
<Geolocation 
    map = { this.map }
    alwaysRecenterToPosition={false}
    colorOrStyleFunc={((feature) => {
       // May be for perf reason do the setState only when geolocationFeature doesn't exist
       this.setState({ geolocationFeature: feature});
       styleLocation.getImage().setRotation(feature.get('rotation') || 0);
       return styleLocation;
    })}
/>
``` 
and in handleorientation do 
```
geolocationFeature.setRotation(compassdir * (2* Math.PI / 360)); // The change event will be triggered automatically
```

Only Improvment we could make for this use case is to add  a feature property so the user can use it's own feature and he doesn't have to do the setState({feature});

candux:
Wow! Thanks a lot @oterral 
I didn't thought this was possible.

PR #548, opened 2020-03-30
[Geolocation] refactor: make point a property and fix function name

candux:
don't trigger useless renders.

I tried to split the functions but wasn't happy with it. Maybe one of you has a better idea

### DIFF
--- a/src/components/Geolocation/Geolocation.js
+++ b/src/components/Geolocation/Geolocation.js
@@ -97,13 +97,7 @@ class Geolocation extends PureComponent {
     if (!geolocation) {
       onError();
     } else if (!active) {
-      this.watch = navigator.geolocation.watchPosition(
-        this.update.bind(this),
-        this.error.bind(this),
-        {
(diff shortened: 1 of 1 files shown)


src/components/Geolocation/Geolocation.js
```diff
@@ -145,10 +143,10 @@ class Geolocation extends PureComponent {
     this.setState({ active: true });
   }
 
-  highlight(point) {
+  highlight() {
     const { colorOrStyleFunc } = this.props;
     const feature = new Feature({
-      geometry: point,
+      geometry: this.point,
       source: new VectorSource(),
```
oterral: this line seems useless.

oterral:
```suggestion
```

candux: good catch!

candux:
@oterral Good idea. I tried to do that

mhaertwig:
@friedjoff can you check, please?

PR #552, opened 2020-04-01
BaseLayerSwitcher t-function

danji90:
- Added t prop to pass in the t-function for translation of layer names
- Texted locally in wkp

### DIFF
--- a/src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
+++ b/src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
@@ -35,6 +35,12 @@ const propTypes = {
     openSwitcher: PropTypes.string,
     closeSwitcher: PropTypes.string,
   }),
+
+  /**
+   * Translation function.
+   * @param {function} Translation function returning the translated string.
+   */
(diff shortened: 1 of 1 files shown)


src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -152,8 +160,8 @@ function BaseLayerSwitcher({
               key={layer.key}
               className={`rs-base-layer-switcher-button ${openClass}${activeClass}`}
               role="button"
-              title={layerName}
-              aria-label={layerName}
+              title={t(layerName)}
```
danji90, line 163: I was debating on whether to apply the t-function when declaring the layerName variable at the beginning of the loop and thereby calling it fewer times, but since I'm using that variable to create the activeClass variable I thought it was a bit dodgy (I would have to translate currentLayer.getName() as well), so I apply it every time I call layerName. Let me know if you agree.

oterral: no do this here is easier

PR #553, opened 2020-04-01
Add position absolute to mapbox canvas.

oterral:
# How to

We don't have mapbox layer on example. So you have to test with `yarn link` in your own project.

# Others

### DIFF
--- a/src/layers/MapboxLayer.js
+++ b/src/layers/MapboxLayer.js
@@ -36,7 +36,7 @@ export default class MapboxLayer extends Layer {
         const canvas = this.mbMap.getCanvas();
         const { viewState } = frameState;
 
-        const visible = mbLayer.getVisible();43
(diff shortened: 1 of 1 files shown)


src/layers/MapboxLayer.js
```diff
@@ -36,8 +36,12 @@ export default class MapboxLayer extends Layer {
         const canvas = this.mbMap.getCanvas();
         const { viewState } = frameState;
 
-        const visible = mbLayer.getVisible();
+        const visible = mbLayer.getVisible();43
```
oterral, line 39:
```suggestion
        const visible = mbLayer.getVisible();
```

candux:
Fixes my problems with mapbox-1.9.0. Before, it didn't show any OL-Layers anymore.
Now everything works. It feels now even snappier https://comforts.space/

PR #555, opened 2020-04-06
KML fillPattern

danji90:
- Added kml writing capability for CanvasPatterns
- When detecting a CanvasPattern in a Fill style  fillPattern: <string> is saved in the feature properties
- This string can be used to create custom CanvasPatterns in the Client app

### DIFF
--- a/src/utils/KML.js
+++ b/src/utils/KML.js
@@ -306,6 +306,12 @@ const writeFeatures = (layer, featureProjection) => {
       }
     }
 
+    if (newStyle.fill && newStyle.fill.getColor() instanceof CanvasPattern) {
+      // eslint-disable-next-line no-console
+      clone.set('fillPattern', f.get('fillPattern'));
+      newStyle.fill = null;
+    }
+
     // If only text is displayed we must specify an
     // image style with scale=0
     if (newStyle.text && !newStyle.image) {
--- a/src/utils/KML.js
+++ b/src/utils/KML.js
@@ -307,7 +307,6 @@ const writeFeatures = (layer, featureProjection) => {
     }
 
     if (newStyle.fill && newStyle.fill.getColor() instanceof CanvasPattern) {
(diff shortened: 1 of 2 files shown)


src/utils/KML.js
```diff
@@ -306,6 +306,12 @@ const writeFeatures = (layer, featureProjection) => {
       }
     }
 
+    if (newStyle.fill && newStyle.fill.getColor() instanceof CanvasPattern) {
+      // eslint-disable-next-line no-console
```
friedjoff, line 310:
```suggestion
```

src/utils/KML.js
```diff
@@ -306,6 +306,12 @@ const writeFeatures = (layer, featureProjection) => {
       }
     }
 
+    if (newStyle.fill && newStyle.fill.getColor() instanceof CanvasPattern) {
```
oterral, line 309: You should exclude also CanvasGradient

oterral: and test if f.get('fillPattern') exists . maybe just a :

``` 
if (newStyle.fill &&  f.get('fillPattern'))
``` 

is enough

danji90: @oterral Did you mean something like this?
`if (
      (newStyle.fill && newStyle.fill.getColor() instanceof CanvasPattern) ||
      newStyle.fill.getColor() instanceof CanvasGradient
    ) {
      if (f.get('fillPattern')) {
        clone.set('fillPattern', f.get('fillPattern'));
        newStyle.fill = null;
      }
    }`

oterral: no just :
``` 
if (newStyle.fill &&  f.get('fillPattern')) {
   clone.set('fillPattern', f.get('fillPattern'));
   newStyle.fill = null;
}
```

oterral:
Perfect thx

PR #557, opened 2020-04-08
Fix listening of resize of the map

oterral:
Reverts geops/react-spatial#556

To test with plan-editor and trafimage-maps before merge.


src/components/BasicMap/BasicMap.js
```diff
@@ -290,18 +304,19 @@ class BasicMap extends PureComponent {
 
   render() {
     const { className, tabIndex, ariaLabel } = this.props;
+    const { node } = this.state;
     return (
       <div
         className={className}
-        ref={this.node}
+        ref={this.setNode}
```
friedjoff, line 311: Shouldn't this by like:
```suggestion
        ref={node}
```

oterral: NO because it comes from the state. so we have to make a setState to update it then trigerrs a re-render

PR #558, opened 2020-04-16
added textAlign to read/Write KML

danji90:
Added textAlign to read/write KML


danji90:
@friedjoff Added textAlign in the test for text style

PR #559, opened 2020-04-21
KML updates

danji90:
- fillPattern attribute now handles all parsing and stringifying in the read/write methods 
- Added offsetX and offsetY parameters to extended data (useful for text features with textAlign attribute)
- Updated tests

### DIFF
--- a/src/utils/KML.js
+++ b/src/utils/KML.js
@@ -125,12 +125,12 @@ const sanitizeFeature = (feature) => {
         text.setTextAlign(feature.get('textAlign'));
       }
 
-      if (feature.get('offsetX')) {
-        text.setOffsetX(parseFloat(feature.get('offsetX')));
+      if (feature.get('textOffsetX')) {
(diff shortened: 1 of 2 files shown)


src/utils/KML.js
```diff
@@ -125,6 +125,14 @@ const sanitizeFeature = (feature) => {
         text.setTextAlign(feature.get('textAlign'));
       }
 
+      if (feature.get('offsetX')) {
```
friedjoff, line 128: How about adding a `text` prefix?
```suggestion
      if (feature.get('textOffsetX')) {
```

friedjoff: Applies also to offsetY.

danji90: You think so? Since the ol attributes are only offsetX and offsetY and there are no other classes using these names, I would use the exact ol Attribute names

danji90: Done

PR #562, opened 2020-04-30
KML zIndex fix

danji90:
The current kml does not properly write the kml index of the feature styles
Fixed by adding zIndex to extended data


danji90:
@friedjoff I always forget the tests :/
Added and ready

PR #563, opened 2020-04-30
Lucien/baselayertogler fix

edellucien:
# How to

# Others

- [x] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/src/components/BaseLayerToggler/BaseLayerToggler.js
+++ b/src/components/BaseLayerToggler/BaseLayerToggler.js
@@ -236,11 +236,10 @@ class BaseLayerToggler extends Component {
     } else {
       newIdx = idx + 1;
     }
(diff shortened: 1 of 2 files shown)


src/components/BaseLayerToggler/BaseLayerToggler.js
```diff
@@ -236,10 +236,11 @@ class BaseLayerToggler extends Component {
     } else {
       newIdx = idx + 1;
     }
-
+    const newLayerVisible = layers[idx === -1 ? 0 : idx];
     this.setState({
       layers,
       idx: layers.length > 1 ? newIdx : null,
+      layerVisible: newLayerVisible,
```
friedjoff, line 243: Would this work as well? That would be the same behavior as in `updateState()`.
```suggestion
      layerVisible: layers.length > 1 ? layers[newIdx] : null,
```

PR #564, opened 2020-04-30
Updates BaseLayerSwitcher

danji90:
Updates:
- Changes layer order on layer switch (puts visible layer at first slot, now assigns images to layer on component mount)
- Updated close button SVG
- Updated focus style


src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable jsx-a11y/interactive-supports-focus */
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { FaChevronCircleLeft } from 'react-icons/fa';
+import ChevronLeft from '../../images/chevron_left.svg';
```
friedjoff, line 4:
```suggestion
import ChevronLeft from '../../images/chevronLeft.svg';
```

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -109,6 +109,15 @@ function BaseLayerSwitcher({
 
   const nextImage = getNextImage(currentLayer, baseLayers, images);
 
+  useEffect(() => {
```
friedjoff, line 112: Why is this in a `useEffect` hook? Why do you rename `baselayer`? Why do you use a `map` function without doing something with the result?

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -143,41 +152,43 @@ function BaseLayerSwitcher({
...
-            >
-              <div className="rs-base-layer-switcher-title">{t(layerName)}</div>
-              {images[index] ? null : (
-                <span className="rs-alt-text">{t(altText)}</span>
-              )}
-            </div>
-          );
-        })
+        baseLayers
+          .sort((l1) => (l1.name === currentLayer.name ? -1 : 0))
```
friedjoff, line 164: This does not seem to work.

danji90: I tested it, it seems to work for me. But maybe it's not the best way to achieve the goal

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: I want to assign the corresponding image to each layer componentDidMount-style, so only the first time the component renders, which is why it is a useEffect with an empty array of dependencies. forEach instead of map would suffice here, I agree

friedjoff: We already map over the baseLayers in line 165 so we can do this stuff directly there. Should be pretty much the same in terms of performance and will be cleaner.

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
friedjoff: If this should move the current selected layer to the beginning of the selection list, it does not work for me. The general approach looks good to me.

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: Done

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: Since .sort() seems to behave slightly differently on different browsers I implemented a different sort method, see updates

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: Removed this hook

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -109,6 +112,14 @@ function BaseLayerSwitcher({
 
   const nextImage = getNextImage(currentLayer, baseLayers, images);
 
+  /* Move visible layer to front of array (array.sort() not working in Firefox) */
```
friedjoff, line 115: Array.sort is an "in place" operation so you shouldn't chain it with a `.map()` call. Can you try to replace this code block with `baseLayers.sort(...)`?

danji90: I now managed to fix sort(), Mozilla's behaviour may turn out inconsistent if the returned value in the compareFunction is 0 (stated in MDN https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort), but in our case we can use 1 instead

PR #566, opened 2020-05-06
Consolidate layer parameters

danji90:
- Removed zIndex, isAlwaysExpandable, previewImage and radioGroup from Layer props
- If required zIndex should now be set using [setZIndex](https://openlayers.org/en/latest/apidoc/module-ol_layer_Layer-Layer.html#setZIndex) on the Layer olLayer
- isAlwaysExpandable, previewImage and radioGroup now need to be specified in the Layer properties
- Updated all services, data and components using the removed parameters
- Parameters can now be accessed using e.g. Layer.get('radioGroup') instead of Layer.radiogroup
- Updated Layer documentation

### DIFF
--- a/src/components/LayerTree/__snapshots__/LayerTree.test.js.snap
+++ b/src/components/LayerTree/__snapshots__/LayerTree.test.js.snap
@@ -366,21 +366,100 @@ exports[`LayerTree matches snapshots using default properties. 1`] = `
           <span />
         </label>
         <div
-          aria-expanded={false}
-          aria-label="Countries Show sublayer"
+          aria-expanded={true}
+          aria-label="Countries Hide sublayer"
           className="rs-layer-tree-toggle"
           onClick={[Function]}
           onKeyPress={[Function]}
           role="button"
           tabIndex={0}
-          title="Countries Show sublayer"
+          title="Countries Hide sublayer"
         >
           <div>
             Countries
           </div>
+        </div>
+      </div>
+      <div>
+        <div
+          className="rs-layer-tree-item "
+          style={
+            Object {
+              "paddingLeft": "60px",
+            }
+          }
+        >
+          <label
+            aria-label="Show layer"
+            className="rs-layer-tree-input rs-layer-tree-input-checkbox rs-checkbox"
+            onKeyPress={[Function]}
+            tabIndex={-1}
+            title="Show layer"
+          >
+            <input
+              checked={false}
+              onClick={[Function]}
+              readOnly={true}
+              tabIndex={-1}
+              type="checkbox"
+            />
+            <span />
+          </label>
           <div
-            className="rs-layer-tree-arrow rs-layer-tree-arrow-collapsed"
-          />
+            aria-expanded={false}
+            aria-label="Countries Borders Show sublayer"
+            className="rs-layer-tree-toggle"
+            onClick={[Function]}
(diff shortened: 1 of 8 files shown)


src/layers/Layer.js
```diff
@@ -57,8 +51,9 @@ export default class Layer extends Observable {
 
     if (this.olLayer) {
       this.olLayer.setVisible(this.visible);
-      if (this.zIndex) {
-        this.olLayer.setZIndex(this.zIndex);
+      if (zIndex) {
+        this.properties.zIndex = zIndex;
```
friedjoff, line 55: Why do we need this twice?

src/components/LayerTree/__snapshots__/LayerTree.test.js.snap
```diff
@@ -366,100 +366,21 @@ exports[`LayerTree matches snapshots using default properties. 1`] = `
           <span />
         </label>
         <div
-          aria-expanded={true}
```
friedjoff, line 369: Why is this snapshot test changing?

src/layers/Layer.js
```diff
@@ -6,47 +6,41 @@ import Observable, { unByKey } from 'ol/Observable';
...
  * @param {Object} [options]
- * @param {string} [options.key] information about the key
- * @param {string} [options.name] The name of the new layer
- * @param {ol.layer} [options.olLayer] the {@link https://openlayers.org/en/latest/apidoc/module-ol_layer_Layer-Layer.html ol/Layer}
- * @param {radioGroup} [options.radioGroup] identifier to group layer in a group, toggle via a radio
- * @param {boolean} [options.isBaseLayer] if true this layer is the baseLayer
- * @param {Array<ol.layer>} [options.children] Layers
- * @param {boolean} [options.visible] If true layer is visible
- * @param {string} [options.copyright] Copyright-Statement
+ * @param {string} [options.key] Layer key.
```
friedjoff, line 9: Please also add documentation about default values and required options.

src/layers/Layer.js
danji90: Mario wants this to be written in the properties so it is accessible via Layer.get('zIndex')

friedjoff: This smells like trouble. What happens if you write `Layer.set('zIndex', 999)`? You end up with two different values.

Why not access this using `Layer.olLayer.getZIndex()`?

danji90: Good point, but your example already applied before my changes

danji90: Will mention this to Mario

src/components/LayerTree/__snapshots__/LayerTree.test.js.snap
danji90: There was a small modification missing in Treedata. Will fix

src/layers/Layer.js
danji90: @friedjoff I set the name and olLayer properties as required, the rest are optional

friedjoff: Please use the [JSDoc syntax for optional and default values](https://jsdoc.app/tags-param.html#optional-parameters-and-default-values).

Sorry, my comment from 4 days ago was a bit misleading. We don't need to specify which parameters are required, just the optional ones using the JSDoc syntax.

PR #569, opened 2020-05-18
BaseLayerSwitcher Animation Update

danji90:
# How to
The BaseLAyerSwitcher opening/closing animation needs to be updated based on the style in https://map.geo.admin.ch/.
Successfully tested locally in Trafimage-Maps 

# Others

- [x] IE11 tested.
- [x] Tests added.


src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -107,7 +107,7 @@ function BaseLayerSwitcher({
     : baseLayers.map((layer) => layer.get('previewImage'));
 
   const openClass = switcherOpen ? ' rs-open' : '';
-  const closedClass = isClosed ? ' rs-closed' : '';
+  const hiddenStyle = switcherOpen && !isClosed ? 'visible' : 'hidden';
```
oterral, line 110: use rs prefix pls

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
```diff
@@ -146,41 +148,58 @@ function BaseLayerSwitcher({
...
-          const layerName = layer.getName();
-          const activeClass =
-            layerName === currentLayer.getName() ? ' rs-active' : '';
-          const imageStyle = getImageStyle(
-            layerImages
-              ? layerImages[`${layer.key}`]
-              : layer.get('previewImage'),
-          );
-          return (
+      {
```
oterral, line 169: uselss I guess

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: Nope, these are not classes, but strings applied to the style object used in react, so adding a prefix will break it

src/components/BaseLayerSwitcher/BaseLayerSwitcher.js
danji90: completely useless

danji90:
> Looks very smooth .
> 
> The last problem I see is the width:600px size of wrapper. You can't move the map. Any chance to calculate this width automatically? or to reduce it? or to allow the mouse events to go to the map?

Yes you're right, I forgot to add this fix. Will try to find a CSS solution

danji90:
@oterral I removed the pointer events from the parent container so they pass through to the map. Please give it a last review and let me know when to merge

gevos:
> I removed the pointer events from the parent container so they pass through to the map. Please give it a last review and let me know when to merge

Looks good.

PR #575, opened 2020-05-27
Don't set a default tabIndex on the BasicMap

edellucien:
# How to
openlayers tabIndex behavior changed a bit due to:
https://github.com/openlayers/openlayers/pull/10291/files

When a tabIndex is set we now have to focus the map to be able to interact with it.
Thus a default map without TabIndex seems to be better.

# Others

- [x] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/src/utils/KML.test.js
+++ b/src/utils/KML.test.js
@@ -42,9 +42,6 @@ describe('KML', () => {
                     <color>ff056600</color>
                     <width>1</width>
                   </LineStyle>
-                  <PolyStyle>
-                    <color>ffffffff</color>
-                  </PolyStyle>
                 </Style>
                 <ExtendedData>
                   <Data name="lineDash"><value>40,40</value></Data>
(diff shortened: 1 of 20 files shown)


src/utils/KML.test.js
```diff
@@ -148,7 +145,7 @@ describe('KML', () => {
 
       // Text
       const style = styles[0].getText();
-      expect(style.getText()).toBe('   bar  ');
+      expect(style.getText()).toBe('bar'); // spaces are trimmed.
```
oterral, line 148: That can be a problem I think plan editor need this.

src/utils/KML.test.js
```diff
@@ -175,6 +172,9 @@ describe('KML', () => {
                         <color>ff0000eb</color>
                         <width>2</width>
                     </LineStyle>
+                    <PolyStyle>
+                      <fill>0</fill>
+                    </PolyStyle>
```
oterral, line 177: We should test in Plan Editor if it breaks something.

PR #581, opened 2020-06-27
fix setting rotation to 0 in MapBox Layer

candux:
# How to

Resetting the rotation to zero with  `this.map.getView().setRotation(0)` did not work in the MapBox-Layer. 
We have to check that `rotation` is not equal to null/undefined instead of false

# Others

- [ ] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/src/layers/MapboxLayer.js
+++ b/src/layers/MapboxLayer.js
@@ -58,7 +58,7 @@ export default class MapboxLayer extends Layer {
 
         // adjust view parameters in mapbox
         const { rotation } = viewState;
-        if (rotation && this.renderState.rotation !== rotation) {
+        if (rotation != null && this.renderState.rotation !== rotation) {
           this.mbMap.rotateTo((-rotation * 180) / Math.PI, {
             animate: false,
           });


src/layers/MapboxLayer.js
```diff
@@ -58,7 +58,7 @@ export default class MapboxLayer extends Layer {
 
         // adjust view parameters in mapbox
         const { rotation } = viewState;
-        if (rotation && this.renderState.rotation !== rotation) {
+        if (rotation != null && this.renderState.rotation !== rotation) {
```
oterral, line 61: it could potentially break if rotation is undefined .
```suggestion
        if (this.renderState.rotation !== rotation) {
```

src/layers/MapboxLayer.js
```diff
@@ -58,7 +58,7 @@ export default class MapboxLayer extends Layer {
...
           this.mbMap.rotateTo((-rotation * 180) / Math.PI, {
```
oterral, line 62:
```suggestion
          this.mbMap.rotateTo((-(rotation || 0) * 180) / Math.PI, {
```

oterral:
Thanks for poiting this out, I propose a diffeent approach to handle null and undefined at the same time

src/layers/MapboxLayer.js
candux: In JS: null == undefined

candux:
> Thanks for poiting this out, I propose a diffeent approach to handle null and undefined at the same time

I think undefined is already handled because null == undefined. If you like your approach still more, I will change it

src/layers/MapboxLayer.js
oterral: you're right but it's weird eslint didn' complain about this. Normally wit should accept only !==

oterral:
Fixed in mobility-toolbox now : https://github.com/geops/mobility-toolbox-js/commit/3104629b98032e57f20f72ac1665521ec161b84d

PR #582, opened 2020-06-28
add hitTolerance on featureClicked

candux:
# How to

clicking on a line-feature on a mobile-phone is really hard. Introduce a hitTolerance property to make it easier.
Default stays the same at 0.

# Others

- [ x Everything in ticket description has been fixed.
- [ ] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/src/components/BasicMap/BasicMap.js
+++ b/src/components/BasicMap/BasicMap.js
@@ -49,6 +49,9 @@ const propTypes = {
    */
   onFeaturesClick: PropTypes.func,
 
+  /** Hit-detection tolerance in pixels. Pixels inside the radius around the given position will be checked for features. */
+  featuresClickHitTolerance: PropTypes.number,
+
   /**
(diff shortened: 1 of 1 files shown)


src/components/BasicMap/BasicMap.js
```diff
@@ -49,6 +49,9 @@ const propTypes = {
    */
   onFeaturesClick: PropTypes.func,
 
+  /** Hit-detection tolerance in pixels. Pixels inside the radius around the given position will be checked for features. */
+  featuresClickHitTolerance: PropTypes.number,
```
oterral, line 53: I guess we could want to pass other options. call it  featuresClickOptions then pass the all object to getFeaturesAtPixel call.

candux: great idea. I will change that

edellucien: https://github.com/candux/react-spatial/pull/1

PR #592, opened 2020-07-07
Filter rendered trajectories by px

friedjoff:
# How to

Only used by trafimage-maps MapAccessability component right now.


friedjoff:
Moved to https://github.com/geops/mobility-toolbox-js/pull/18

PR #605, opened 2020-08-27
BLS Switcher title update

danji90:
# How to

When only two base maps are present, the layer name is shown on the switcher opener directly (otherwise it shows a configurable title like "Switch Baselayer")

Test:

- Load the review app styleguide
- In the example for the BaseLayerSwitcher, in the BaseLayerSwitcher props change layers={layers} to layers={layers.slice(1)} (remove one base layer to a total of two)
- Click on the BaseLayerSwitcher
- The Switcher should toggle the base map instead of opening the Switcher
- Switcher title should be Layer names

# Others

- [x] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-spatial",
   "description": "Components to build React map apps.",
-  "version": "1.0.4-beta.1",
+  "version": "1.0.4-beta.2",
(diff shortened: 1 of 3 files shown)


friedjoff (approved):
LGTM

PR #612, opened 2020-09-07
KML icon rotation

danji90:
# How to

In mapset, it should be possible to rotate icons with the feature styler.
Currently this does not work, since the "rotation" value in icon styles is written as a <heading> tag into the kml string,
which in turn is not read by the KML module and the value isn't added to any feature property by default. 

Solution: We write the value into the kml extended data and read it from there when loading rotated icons 

# Others

- [] IE11 tested.
- [] Labels applied. if it's a release? a hotfix?
- [x] Tests added.

### DIFF
--- a/src/utils/KML.js
+++ b/src/utils/KML.js
@@ -164,6 +164,11 @@ const sanitizeFeature = (feature) => {
       }
     }
 
+    if (image instanceof Icon) {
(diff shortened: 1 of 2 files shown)


friedjoff (approved):
LGTM

👍  for adding a test!

PR #631, opened 2020-11-05
Use StopAPI from mobility-toolbox-js in StopFinder

edellucien:
# How to

Use StopAPI from mobility-toolbox.js

# Others

- [ ] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/src/components/Search/Search.js
+++ b/src/components/Search/Search.js
@@ -51,7 +51,7 @@ const propTypes = {
 
 const defaultProps = {
   apiKey: null,
-  engines: { stops: new StopFinder() },
+  engines: null,
   getRenderSectionTitle: () => () => null,
   onHighlight: () => null,
   shouldRenderSuggestions: (newValue) => newValue.trim().length > 2,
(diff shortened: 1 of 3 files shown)


src/components/Search/engines/StopFinder.js
```diff
@@ -1,25 +1,22 @@
...
-    return fetch(
-      `${this.endpoint}?&q=${encodeURIComponent(value)}&key=${
-        this.apiKey
-      }&${optionsString}`,
-    )
-      .then((data) => data.json())
-      .then((featureCollection) => featureCollection.features);
+    return this.api.search({
+      q: value,
+    });
```
oterral, line 19: What happens with the use of options and the encodeURIComponnet function ? If they were there it was for a resaon I guess

oterral (changes requested):
Did you test with trafimage  and other ?

src/components/Search/engines/StopFinder.js
edellucien: Indeed forgot that! Thanks

edellucien: If that is it, we could merge it.

src/components/Search/Search.js
```diff
@@ -70,11 +70,21 @@ function Search({
...
   className,
 }) {
+  const currentEngines = useMemo(() => {
+    if (!engines) {
+      return {
+        stops: new StopFinder(null, { apiKey }),
+      };
+    }
+    return engines;
+  }, [apiKey, engines]);
```
oterral, line 80: Does this code could be in the useMemo of SearchService ?

src/components/Search/engines/StopFinder.js
oterral: and options ?

edellucien: Of course you are right, options are added as well

oterral (approved):

PR #633, opened 2020-11-11
Lucien/use stop api

edellucien:
# How to

Fix Use of Search with engines & apiKey prop
# Others

- [ ] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/src/components/Search/README.md
+++ b/src/components/Search/README.md
@@ -26,7 +26,6 @@ const layers = [
 // Please get your own api key at https://developer.geops.io/.
 const { apiKey } = window;
 
-
 <div className="rt-stop-finder-example">
   <BasicMap
     map={map}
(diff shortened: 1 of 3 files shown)


friedjoff (changes requested):
Doesn't seem to use the StopAPI from mobility-toolbox-js anymore.

edellucien:
In StopFinder we are loading:
`import { StopsAPI } from 'mobility-toolbox-js/api';`

danji90 (changes requested):
@edellucien The requests in the rs-Demo now send requests with two key parameters (_apiKey_ and _key_)

edellucien:
Thanks! Duplicates is created due to StopsAPI. I delete apiKey when from options passed to 'StopsAPI' to avoid this duplicate.

friedjoff (approved):

PR #638, opened 2020-11-25
StopFinder component

oterral:
# How to

- Open StopFinder example and use it

# Others

- [ ] IE11 tested.
- [ ] Tests added.


package.json
```diff
@@ -1,12 +1,13 @@
...
   "name": "react-spatial",
   "description": "Components to build React map apps.",
-  "version": "1.0.13",
+  "version": "1.0.14-beta.4",
   "license": "MIT",
   "dependencies": {
     "@geops/geops-ui": "0.1.8",
     "@material-ui/core": "^4.11.0",
     "@material-ui/icons": "^4.9.1",
+    "@material-ui/lab": "^4.0.0-alpha.56",
```
friedjoff, line 10: Do we want to use material-ui in react-spatial?

friedjoff (commented):
Code looks good in general!

package.json
oterral: Since we use it in all our app it makes sense to have the default component using MUI. @mhaertwig ?
But it's important to have the possibility to use something else if needed.

oterral: we could also use the MUI autocomplete only in the example.

PR #649, opened 2021-01-22
Permalink round zoom

edellucien:
Ensure the zoom in permalink can only have 2 digit max for readability

- [ ] IE11 tested.
- [ ] Tests added.

### DIFF
--- a/src/components/Permalink/Permalink.js
+++ b/src/components/Permalink/Permalink.js
@@ -66,6 +66,11 @@ const defaultProps = {
  * be added using __params__.
  */
 class Permalink extends Component {
+  static roundToTwo(num) {
+    // rounds number to two digits max if necessary.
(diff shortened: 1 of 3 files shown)


src/components/Permalink/Permalink.js
```diff
@@ -147,7 +152,7 @@ class Permalink extends Component {
     this.setState({
       x: this.roundCoord(center[0]),
       y: this.roundCoord(center[1]),
-      z: mapView.getZoom(),
+      z: Permalink.roundToTwo(mapView.getZoom()),
```
mhaertwig, line 155: The function just reduces readability. Do the rounding here, please.

edellucien: I removed the function as you prefer

PR #673, opened 2021-09-15
KML: Add methods for writing, removing and reading <Camera> tags

danji90:
# How to
Updates:
- readDocumentCamera
- writeDocumentCamera
- removeDocumentCamera

The new methods handle the reading and writing of <Camera> tags in the KML (contain data for View rotation, tilt etc).
See https://developers.google.com/kml/documentation/cameras?hl=en for details.

# Others

- [x] Tests added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-spatial",
   "description": "Components to build React map apps.",
-  "version": "1.0.31-beta.1",
+  "version": "1.0.31-beta.2",
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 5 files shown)


oterral (changes requested):
I would prefer or more openlayers centric approach , try to implement a KML format class that extends the openlayers format class. Take example  the readRegionXXX function in ol.format.KML. So we cna move it to OL afterwards

danji90:
@oterral PR in openlayers: https://github.com/openlayers/openlayers/pull/12787

PR #680, opened 2021-12-22
Update mobility-toolbox-js@1.6.3

oterral:
# How to

This PR fix imports to prepare to use it with the last version of react-scripts  and change the RouteSchedule example

# Others

- [x] Tests added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-spatial",
   "description": "Components to build React map apps.",
-  "version": "1.0.36",
+  "version": "1.0.37-beta.1",
   "license": "MIT",
   "dependencies": {
     "@geops/geops-ui": "0.1.12",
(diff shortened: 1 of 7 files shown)


danji90 (changes requested):
@oterral There are still a couple of issues:
- The route schedule doesn't close when the a route is deselected via map click (if this is the desired behaviour we should at least add a close button on the schedule container)
- Sometimes when you switch from one route to another residual stations from the last schedule are rendered above the Perlenkette. I dunno if this bug was already there before though

oterral:
> @oterral There are still a couple of issues:
> 
> * The route schedule doesn't close when the a route is deselected via map click (if this is the desired behaviour we should at least add a close button on the schedule container)
Fixed
> * Sometimes when you switch from one route to another residual stations from the last schedule are rendered above the Perlenkette. I dunno if this bug was already there before though
Fixed but there i also a backend problem. I have contacted the devs.

PR #697, opened 2022-11-02
fix(LayerTree): make LayerTree renderItemContent more configurable

danji90:
# How to

# Others

- [ ] The title means something for a human being and follows the [conventional commits](https://www.conventionalcommits.org/) specification.
- [ ] The title contains [WIP] if it's necessary.
- [ ] Tests added.

### DIFF
--- a/src/components/LayerTree/LayerTree.js
+++ b/src/components/LayerTree/LayerTree.js
@@ -44,8 +44,11 @@ const propTypes = {
 
   /**
    * Custom function to render only the content of an item in the tree.
+   * inputProps und toggleProps can be used when calling the default renderItemContent function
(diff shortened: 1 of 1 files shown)


src/components/LayerTree/LayerTree.js
```diff
@@ -317,7 +317,7 @@ class LayerTree extends Component {
         aria-label={layer.visible ? titles.layerHide : titles.layerShow}
         onKeyPress={(e) => {
           if (e.which === 13) {
-            this.onInputClick(layer);
+            this.onInputClick(layer, undefined, inputType);
```
oterral, line 320: Why do we need this ?

danji90: We don't

PR #723, opened 2024-06-12
fix(onButtonClick): add onButtonClick callback function prop to BaseLayerSwitcher and Zoom

danji90:
# How to
In WKP we need to track certain events for the SBB, including the BaseLayerSwitcher and Zoom buttons.
A prop was added to trigger a callback function on click of the buttons in the components

# Others

- [ ] The title means something for a human being and follows the [conventional commits](https://www.conventionalcommits.org/) specification.
- [ ] The title contains [WIP] if it's necessary.
- [ ] Tests added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "react-spatial",
   "license": "MIT",
   "description": "Components to build React map apps.",
-  "version": "1.11.1-beta.1",
+  "version": "1.11.1-beta.2",
   "dependencies": {
     "@emotion/react": "^11.11.4",
     "@emotion/styled": "^11.11.5",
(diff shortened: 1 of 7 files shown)


oterral (changes requested):
I prefer that you add multiple onXXXClick prop instead of one that you called everywhere.

It will avoid to maage the 2nd parameter which likely is only useful in application, totally useless in react-spatial.

onCloseButttonClick
onLayerButtonClick
onSwitcherButtonClick

onZoomInButtonClick
onZoomOutButtonClick

danji90:
> I prefer that you add multiple onXXXClick prop instead of one that you called everywhere.
> 
> It will avoid to maage the 2nd parameter which likely is only useful in application, totally useless in react-spatial.
> 
> onCloseButttonClick onLayerButtonClick onSwitcherButtonClick
> 
> onZoomInButtonClick onZoomOutButtonClick

makes sense. Done

oterral (approved):

